PR #150, opened 2018-09-27
A few small changes for SFR support for MIRS

spruceboy:
Dave, here are a few small changes we made for mirs support. The scaling needs to be confirmed with the mirs folks, we are sort of unclear what the maximum value is.   Feel free to discard these changes if they don't fit in your plans, just wanted to let you know what we did. Thanks! 

Oh, I am going to create a follow up ticket with another issue with no data values.

### DIFF
--- a/polar2grid/core/rescale_configs/rescale.ini
+++ b/polar2grid/core/rescale_configs/rescale.ini
@@ -181,11 +181,10 @@ min_in=0.0
 max_in=50.0
 
 [rescale:default_sfr]
-data_kind=sfr
+data_kind=snow_fall_rate
 method=linear
 min=0.0
-max_in=50.0
-#This should probibly be from 0 to 5, or some other max value, pending a conversation with the mirs team.
+max_in=5.0
 
 [rescale:default_optical_depth]
 data_kind=optical_depth
--- a/polar2grid/core/rescale_configs/rescale.ini
+++ b/polar2grid/core/rescale_configs/rescale.ini
@@ -180,6 +180,13 @@ method=linear
 min_in=0.0
 max_in=50.0
 
+[rescale:default_sfr]
+data_kind=sfr
+method=linear
+min=0.0
+max_in=50.0
+#This should probibly be from 0 to 5, or some other max value, pending a conversation with the mirs team.
+
 [rescale:default_optical_depth]
 data_kind=optical_depth
 method=linear
--- a/polar2grid/mirs/mirs2swath.py
+++ b/polar2grid/mirs/mirs2swath.py
@@ -207,6 +207,7 @@ TPW_VAR = "tpw_var"
 SWE_VAR = "swe_var"
 CLW_VAR = "clw_var"
 TSKIN_VAR = "tskin_var"
+SFR_VAR = "sfr_var"
 
 PRODUCT_RAIN_RATE = "rain_rate"
 PRODUCT_BT_CHANS = "btemp_channels"
@@ -219,6 +220,7 @@ PRODUCT_TPW = "tpw"
 PRODUCT_SWE = "swe"
 PRODUCT_CLW = "clw"
 PRODUCT_TSKIN = "tskin"
+PRODUCT_SFR = "sfr"
 
 PAIR_MIRS_NAV = "mirs_nav"
 
@@ -234,6 +236,7 @@ PRODUCTS.add_product(PRODUCT_TPW, PAIR_MIRS_NAV, "total_precipitable_water", FT_
 PRODUCTS.add_product(PRODUCT_SWE, PAIR_MIRS_NAV, "snow_water_equivalence", FT_IMG, SWE_VAR, description="Snow Water Equivalence", units="cm")
 PRODUCTS.add_product(PRODUCT_CLW, PAIR_MIRS_NAV, "cloud_liquid_water", FT_IMG, CLW_VAR, description="Cloud Liquid Water", units="mm")
 PRODUCTS.add_product(PRODUCT_TSKIN, PAIR_MIRS_NAV, "skin_temperature", FT_IMG, TSKIN_VAR, description="skin temperature", units="K")
+PRODUCTS.add_product(PRODUCT_SFR, PAIR_MIRS_NAV, "snow_fall_rate", FT_IMG, SFR_VAR, description="snow fall rate", units="mm/hr")
 
 
 GEO_PAIRS = GeoPairDict()
@@ -254,6 +257,7 @@ FILE_STRUCTURE = {
     SWE_VAR: ("SWE", ("scale", "scale_factor"), None, None),
     CLW_VAR: ("CLW", ("scale", "scale_factor"), None, None),
     TSKIN_VAR: ("TSkin", ("scale", "scale_factor"), None, None),
+    SFR_VAR: ("SFR", ("scale", "scale_factor"), None, None),
     }
 
 LIMB_SEA_FILE = os.environ.get("ATMS_LIMB_SEA", "polar2grid.mirs:limball_atmssea.txt")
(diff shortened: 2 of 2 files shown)


polar2grid/core/rescale_configs/rescale.ini
```diff
@@ -180,6 +180,13 @@ method=linear
 min_in=0.0
 max_in=50.0
 
+[rescale:default_sfr]
+data_kind=sfr
+method=linear
+min=0.0
+max_in=50.0
+#This should probibly be from 0 to 5, or some other max value, pending a conversation with the mirs team.
+
```
djhoese: @kathys Do you think you could contact the MIRS team about this and see if snow fall rate is really only useful between 0 and 5 mm/hr?

polar2grid/core/rescale_configs/rescale.ini
```diff
@@ -180,6 +180,13 @@ method=linear
 min_in=0.0
 max_in=50.0
 
+[rescale:default_sfr]
+data_kind=sfr
```
djhoese: I think this needs to be "snow_fall_rate". Otherwise you could add `product_name: sfr` which would accomplish the same thing in this case. What we want is to configure this to match any product where `data_kind` equals `snow_fall_rate` (as configured in the reader) or `product_name` equals `srf` (as named in the reader). You don't need both.

polar2grid/core/rescale_configs/rescale.ini
djhoese: @kathys I forget, did we ever get a response on this?

kathys: On September 28, 2018, Jim Davies sent this to me from Chris Grassotti:
We were contacted directly by the Alaska folks about the SFR and gave them some guidance. The max value in the SFR (after applying the scale factor of 100 in the netcdf) is about 5 mm/h (stored as 500 in the file). The maximum possible value of 1000 in the metadata was a sort of placeholder and will probably be revised on a future release of MiRS.

polar2grid/core/rescale_configs/rescale.ini
spruceboy: I tried that, and the log file implies it is doing what it should:
```
[2019-03-27 19:27:18,383] : PID  30408 : DEBUG    : polar2grid.core.rescale : get_rescale_options : Product sfr found in rescale config: {'satellite': 'noaa19', 'units': 'mm/hr', 'data_type': 'uint1', 'min': '0.0', 'reader': 'mirs', 'product_name': 'sfr', 'instrument': 'amsua-mhs', 'grid_name': '203_10km', 'max_in': 5.0, 'inc_by_one': False, 'data_kind': 'snow_fall_rate', 'method': 'linear'}
```

But it still doesn't look correct.  I will take to Carl and see what he says.

djhoese: @spruceboy was AWIPS updated to the new 0-5 range?

spruceboy: Hmm, good point, let me check with Carl.

spruceboy: Hmm, #2 - looking at the raw netcdf output of polar2grid, it looks like most of the land is masked off with the value of "6", and no snow is masked off with a value of 1.  
Is there some sort of special value in this product for invalid values? 

Here is the data I am looking at:
http://hippy.gina.alaska.edu/distro/tmp/waffle/

spruceboy:
It looks like there is a separate invalid data value for land - this is
what the raw netcdf files look like.
Is this somehow accounted for in the SCMI writer?

polar2grid/core/rescale_configs/rescale.ini
spruceboy: This is what the netcdf source data looks like.  The blue is I think some sort of invalid data.. 
Is this handled somehow in the scmi writer so we are not noticing it? 

Screen Shot 2019-03-27 at 12 02 48 PM

djhoese:
The best I can think of right now is maybe the SCMI writer is clipping the data in a "better" way than the legacy AWIPS writer is. Still investigating.

polar2grid/core/rescale_configs/rescale.ini
djhoese: @spruceboy Could you please push your changes to rescale.ini so I can see/use them.

spruceboy: It looks like those blue values are 0.0  - perhaps the SCMI writer is treating those as nodata, which would hid this problem.

djhoese: Also it looks like you probably have some custom changes to `awips_backend.ini`. What does that look like?

djhoese:
According to the global attributes in the file you linked to there are 3 fill values:

```
                :missing_value = -999 ;
                :notretrievedproduct_value = -888 ;
                :noretrieval_value = -99 ;
```

I think we're only checking for the single missing_value right now. If I load the data after unscaling it from the NetCDF file but before resampling I get these unique values:

```
In [10]: np.unique(data)                                                                                                                                                         
Out[10]: 
array([-0.13      , -0.09999999,  0.        ,  0.04      ,  0.05      ,
        0.06      ,  0.07      ,  0.08      ,  0.09      ,  0.09999999,
        0.11      ,  0.12      ,  0.13      ,  0.14      ,  0.14999999,
        0.16      ,  0.17      ,  0.17999999,  0.19      ,  0.19999999,
        0.22      ,  0.28      ,  0.31      ,  0.39      ], dtype=float32)
```

Without unscaling the data (raw bytes in the data) I see:

```
In [9]: np.unique(nc['SFR'][:])                                                                                                                                                  
Out[9]: 
array([-13, -10,   0,   4,   5,   6,   7,   8,   9,  10,  11,  12,  13,
        14,  15,  16,  17,  18,  19,  20,  22,  28,  31,  39], dtype=int16)
```

Still looking in to things.

djhoese:
Here is the raw swath data from the NetCDF when I look at it (not ignoring fill values):

image

So while the reader isn't doing anything to handle the other missing values, I don't see any in this data either. The majority of the ocean data seems to be -10 (scaled in NetCDF) meaning -0.1 when unscaled.

djhoese:
Another option...you could add `offset=1` to the `rescale.ini` entry which would increment everything by 1 and make 0 the fill value. Right now I think 0 represents both valid values (0) and invalid values due to the way AWIPS 1 used to handle data.

djhoese:
Hhhhmmm actually `offset` may be a bad idea. I don't think the legacy AWIPS writer encodes that in any way so values come out incorrect.

djhoese:
This is what my SCMI file looks like:

image

djhoese:
So how do we know that the SCMI output is correct? From what I'm seeing it seems like the SCMI is wrong and the legacy AWIPS is correct. By that I mean that the input data seems to have actual values (even though some are negative) and these aren't showing up in your version of the SCMI writer but seem to be showing up in mine (at least with my basic tests.

I'm a little confused by how you got that panoply screenshot, wheres all the other data values?

spruceboy:
Let me see if I can get carl to chime in - he has been looking at these a lot, and has some idea of what "correct" is.

djhoese:
@spruceboy I was going to push the fixes directly to your branch which is normally possible, but I think since you made your pull request from your master branch I don't have those permissions. No big deal. I made a temporary branch here: https://github.com/ssec/polar2grid/tree/add-sfr-legacy
But the important part is this commit: https://github.com/ssec/polar2grid/commit/77bc72a9d120795fb792aff3722c6734f3cbfed5

The description of these changes is in #167. I'm not sure why the SCMI writer was "working", but I think you should fix that writer too until a new version of Polar2Grid can be released. The changes to make are in this commit: https://github.com/ssec/polar2grid/commit/d494596d95457796fea9323b69ddea436e7ccea1

I'll send an email specifically to Carl about this, but let me know if this helps.

spruceboy:
Thanks Dave!  I will check with Carl and see if he wants me to make these changes for SCMI on our systems.  

My apologies for making a hard to murge pull request. 

Thanks again for taking the time to look into and fix this problem, I release it is a small one-off case for something that doesn't get a lot of use.

djhoese:
The code for this PR was merged in #167

PR #171, opened 2019-05-24
Add palettize rescaling option to colormap data in one step

djhoese:
In preparation of colormapping VIIRS Active Fires output, this PR adds a new `palettize` rescaling method where a colormap can be provided and applied in the main polar2grid call.

This is a starting point and we'll need to make some decisions before things are finalized. The main issue is that this feature and where I'd like to go with Polar2Grid conflict with previous versions of Polar2Grid. Right now AMSR2 L1b is the only thing users are shown how to colormap (to duplicate NRL's displays). However, these colormaps use `0` as a fill value with a color of black. In future versions (and the current default in Satpy) I'd like images to be generated with an Alpha band to represent transparency. My solution to this is that I've made this PR allow for both possibilities when the proper parameters are specified. To use the legacy-style colormaps:

```
[rescale:amsr2_btemp_36.5v]
data_kind=toa_brightness_temperature
instrument=amsr2
product_name=btemp_36.5v
method=palettize
min_in=180
max_in=280
colormap=$POLAR2GRID_HOME/colormaps/amsr2_36h.cmap
alpha=False
```

To use a colormap from `trollimage` or some other colormap that doesn't have a fill value color:

```
[rescale:confidence_cat]
product_name=confidence_cat
method=palettize
colormap=ylorrd
min_in=7
max_in=9
alpha=True
```

The next issue is what is the default behavior for AMSR-2 L1B data. By default in current versions of Polar2Grid you get a grayscale image with a special BT enhancement. I can make the default add the colormap that *most* people are probably using, but it makes it difficult for users who wanted the grayscale image (custom rescale configs, etc). I could go a step further and edit the builtin colormaps to not have a fill value color in them and make them produce images with an Alpha channel. Here is an example VIIRS I04 band with a 3-color colormap added to it and an alpha band created:

image

Edit/Note: This code is not pretty and does some really hacky things, but in P2G 3.0 this will all get replaced and I won't worry so much about backwards compatibility.


djhoese:
As discussed in our last meeting we will leave the defaults as is since the next release is a minor release (2.3 instead of 3.0). However, the fire products, being new functionality, will use the "alpha" functionality to make invalid/empty pixels transparent. This should make it easier/more logical to merge with RGBs in later documentation...I hope.

PR #172, opened 2019-05-30
Add VIRR_l1b, MERSI2_l1b, VIIRS_edr_flood, and VIIRS_edr_active_fires readers

wroberts4:
Each of the four readers above has a wrapper class that calls their respective readers in Satpy.

Requested and helped by @djhoese.


setup.py
```diff
@@ -158,6 +158,10 @@ def readme():
         'crefl_sharpen_awips=polar2grid.compositors.rgb:CreflRGBSharpenCompositor',
         ],
     'polar2grid.frontend_class': [
+        'viirs_edr_flood=polar2grid.readers.mersi2_l1b:Frontend',
+        'viirs_edr_active_fires=polar2grid.readers.virr_l1b:Frontend',
```
djhoese: Copy/paste error, wrong reader. Same for the frontend arguments below for a couple of them.

polar2grid/readers/mersi2_l1b.py
```diff
@@ -0,0 +1,37 @@
+import sys
+
+import logging
+from polar2grid.readers import ReaderWrapper, main
+
+LOG = logging.getLogger(__name__)
+
+
+class Frontend(ReaderWrapper):
+    FILE_EXTENSIONS = ['.h5']
```
djhoese: This file extension has to match what the actual files will be. For mersi2 this is ".HDF". See https://github.com/pytroll/satpy/blob/master/satpy/etc/readers/mersi2_l1b.yaml

Please double check the other readers too.

polar2grid/readers/viirs_edr_active_fires.py
```diff
@@ -0,0 +1,37 @@
...
+                       help="List available frontend products and exit")
+    group_title = "Frontend Swath Extraction"
+    group = parser.add_argument_group(title=group_title, description="swath extraction options")
+    group.add_argument("-p", "--products", dest="products", nargs="+", default=None, action=ExtendAction,
+                       help="Specify frontend products to process")
+    return ["Frontend Initialization", "Frontend Swath Extraction"]
+
+
+if __name__ == "__main__":
+    sys.exit(main(description="Extract VIRR_EDR_ACTIVE_FIRES swath data into binary files",
```
djhoese: VIIRS not VIRR

djhoese:
The filenames in the docstrings aren't really human friendly, but we'll fix those later.

PR #181, opened 2019-06-25
Fix ll2cr cols/rows not supporting 64-bit floats

wroberts4:
Currently, using libraries that return float64 as the data type for arrays causes polar2grid to stall during cython code execution. This fix allows polar2grid to take more than just float32 as a data type.

Solved by @djhoese.


djhoese:
For the record, what is happening is that ll2cr is writing column and row values to a flat binary file with the same data type as the longitude and latitude arrays. However, the code that reads these files to pass them to fornav was hardcoded to 32-bit floats. This was never a problem before because we always had 32-bit float lon/lat arrays.

PR #188, opened 2019-07-18
Add satellite name normalization for NOAA-NN and SNPP satellites

djhoese:
The various VIIRS readers (crefl, viirs l1b, viir sdr) produce different names for the satellite/platform name that goes in the resulting output filename. This PR normalizes all of them to use "npp" for Suomi-NPP and "noaa20" for NOAA-20/JPSS-1. It should be relatively flexible and work with future noaa satellites following this naming structure.


7yl4r:
My only comment is that I might prefer `normalize_satellite_name` to be structured as an if/elif/else, but that's a very minor personal preference and it is good as-is.

djhoese:
Yeah I thought about it, but given the `return` statements it doesn't really make a functional difference.

Issue #154, opened 2019-02-15
Switch to versioneer for internal python versions


djhoese:
SatPy and other pytroll packages have switched to using the [versioneer](https://github.com/warner/python-versioneer) tool for identifying the version of a python package. This tool lets you "tag" a release in git which will automatically make the python version `python -c 'import polar2grid; print(polar2grid.__version__)'` match the name of the last git tag. This goes further where if the installed version isn't tagged (v3.0.0) then it gets the git commit hash added to the end of it (v3.0.0+abcdefg). This is extremely helpful in identifying if one installed version is the same as a released package.

djhoese:
See https://github.com/pytroll/satpy/issues/678 for details but it may be better to use `setuptools-scm` instead.

PR #189, opened 2019-07-18
Update and fix automated tests with Jenkins

wroberts4:
This is a continuation off of @katherinekolman's work.

2 major things happen:
    1) Runs tests on different aspects of polar2grid.
    2) On success, makes the docs and softwarebundle, then saves them to Bumi.


integration_tests/features/data_conversion.feature
```diff
@@ -7,40 +7,40 @@ Feature: Testing data conversions
...
-
-    Examples: AVHRR
-        | source            | command             | output             | script        |
-        | avhrr/input/test1 | avhrr gtiff -vv -f  | avhrr/output/test1 | polar2grid.sh |                               
-  
-    Examples: MODIS
-        | source            | command                                                                                                                               | output             | script        |
-        | modis/input/test1 | crefl gtiff -vv --true-color --false-color --fornav-D 10 --grid-configs ${POLAR2GRID_HOME}/grid_configs/grid_example.conf -g miami -f | modis/output/test1 | polar2grid.sh |
-
+#    Examples: ABI
```
djhoese: Why are these all commented out?

integration_tests/run.sh
```diff
@@ -1,14 +1,35 @@
...
-export POLAR2GRID_HOME="$WORKSPACE/test_swbundle"
+time=`date +"%Y%m%d-%H%M%S"`
+tarball_name=polar2grid-swbundle-$time
+./create_conda_software_bundle.sh $WORKSPACE/$tarball_name
+export POLAR2GRID_HOME="$WORKSPACE/$tarball_name"
 cd "$WORKSPACE/integration_tests"
 behave --no-logcapture --no-color --no-capture -D datapath=/data/users/kkolman/integration_tests/polar2grid/integration_tests/p2g_test_data
-exit $?
+# Only run by Jenkins if build was successful.
+if [ $? = 0 ]
```
djhoese: I believe this should be `[ $? -eq 0 ]`. Correct me if I'm wrong.

djhoese:
Do you think we can add the handling of whatever jenkins environment variable might hold the current git tag? Or maybe we use `git describe` to see if there's a tag (preference on a jenkins env var)?

Maybe we should come up with a scheme for `g2g-vX.Y.Z` and `p2g-vX.Y.Z` to match the release version?

I was also looking at #154 about automating the python package version number based on a tag, but that will be harder since the python package version doesn't current match the software bundle release. 🤔

integration_tests/features/data_conversion.feature
wroberts4: To save time; these tests take a long time to run, and I am working with pre and post build actions. It's temporary.

integration_tests/run.sh
```diff
@@ -1,14 +1,36 @@
...
+# Script for jenkins to run tests on polar2grid.
 
+export PATH="/usr/local/texlive/2019/bin/x86_64-linux":$PATH
 cd "$WORKSPACE"
-mkdir -p integration_tests/jenkins_p2g_env
+# Activate conda for bash.
+/data/users/davidh/miniconda3/bin/conda init bash
+# Restart the shell to enable conda.
+source ~/.bashrc
+conda remove -y -n jenkins_p2g_env --all
```
djhoese: Why do you remove the environment first? I suppose it doesn't work consistently otherwise?

My hope was that the `conda env update` below would do whatever was necessary to update the existing environment. Based on what I know about conda the way you are doing it now is probably actually faster (because of how the package dependencies get resolved).

wroberts4: I won't need to remove it every time, there were conda-pack errors and I wasn't sure if the environment was corrupted at some point. Removing the environment fixes this temporarily, but I will find a long term fix.

PR #190, opened 2019-07-29
Use Git Tags to Enhance Jenkins Tests

wroberts4:
Jenkins now looks at the git tag (if one is provided) and will make a release of polar2grid or geo2grid based on the git tag name. Git tag naming should follow: package_name-vX.X.X. The package name is one of p2g or g2g (for polar2grid and geo2grid respectively). The version is only for releases; note that if no version is provided, then the package specified will be made using test naming (date at the end). Regardless, on successful build of both, the software-bundle and the documentation will be copied to /tmp.

When a release is made, the software-bundle will be named package_name-swbundle-vX.X.X, and will be included in the directory package_name-vX.X.X along with the html and pdf docs. In such a case, this directory will also be copied to /data/tmp. Otherwise, if a version is not specified, the software-bundle will be named package_name-swbundle-YYYYmmdd-HHMMSS, and will be included in the directory package_name-YYYYmmdd-HHMMSS along with the html and pdf docs. If anything goes wrong with reading the git tag, polar2grid-YYYYmmdd-HHMMSS is used.

Examples:
"git tag p2g-v2.4.3a" -> polar2grid-swbundle-v2.4.3a
"git tag g2g-v1.5.2" -> geo2grid-swbundle-v1.5.2
"git tag g2g" -> geo2grid-swbundle-20190723-144830
"git tag 2g-v1.0.0" -> polar2grid-swbundle-20190723-144830


djhoese:
Replaced by #191

PR #191, opened 2019-07-29
Use Git Tags to Enhance Jenkins Tests

djhoese:
Replace #190 due to git ugliness

Jenkins now looks at the git tag (if one is provided) and will make a release of polar2grid or geo2grid based on the git tag name. Git tag naming should follow: package_name-vX.X.X. The package name is one of p2g or g2g (for polar2grid and geo2grid respectively). The version is only for releases; note that if no version is provided, then the package specified will be made using test naming (date at the end). Regardless, on successful build of both, the software-bundle and the documentation will be copied to /tmp.

When a release is made, the software-bundle will be named package_name-swbundle-vX.X.X, and will be included in the directory package_name-vX.X.X along with the html and pdf docs. In such a case, this directory will also be copied to /data/tmp. Otherwise, if a version is not specified, the software-bundle will be named package_name-swbundle-YYYYmmdd-HHMMSS, and will be included in the directory package_name-YYYYmmdd-HHMMSS along with the html and pdf docs. If anything goes wrong with reading the git tag, polar2grid-YYYYmmdd-HHMMSS is used.

Examples:
"git tag p2g-v2.4.3a" -> polar2grid-swbundle-v2.4.3a
"git tag g2g-v1.5.2" -> geo2grid-swbundle-v1.5.2
"git tag g2g" -> geo2grid-swbundle-20190723-144830
"git tag 2g-v1.0.0" -> polar2grid-swbundle-20190723-144830


integration_tests/run.sh
```diff
@@ -11,27 +11,43 @@ source ~/.bashrc
...
 # Save software bundle.
-rm -rf /tmp/polar2grid-*
-mkdir /tmp/polar2grid-$time
-cp -r "$WORKSPACE/$tarball_name" /tmp/polar2grid-$time
+rm -rf /tmp/${prefix}2grid-*
+mkdir /tmp/${prefix}2grid-$end
+cp -r "$WORKSPACE/$tarball_name" /tmp/${prefix}2grid-$end
 # Make docs.
-conda install -y sphinx
+conda update -c conda-forge -n jenkins_p2g_env sphinx
```
djhoese: Should `sphinx` get added to the `build_environments.yml`? This command will fail if `sphinx` isn't installed in the environment already.

integration_tests/run.sh
```diff
@@ -11,27 +11,43 @@ source ~/.bashrc
...
 # Needs to be second since Jenkins makes an html in workspace from the file generated by this command.
-make html
-cp -r "$WORKSPACE"/doc/build/html /tmp/polar2grid-$time
-chmod a+rx /tmp/polar2grid-$time
\ No newline at end of file
+make html POLAR2GRID_DOC=$prefix
+cp -r "$WORKSPACE"/doc/build/html /tmp/${prefix}2grid-$end
+chmod a+rx /tmp/${prefix}2grid-$end
+if [[ $end = "$version" ]]; then
+    cp -r /tmp/${prefix}2grid-$end /data/tmp/${prefix}2grid-$end
```
djhoese: If the build is versioned and tests succeeded could we copy the resulting `.tar.gz` file to `/data/dist`?

build_environment.yml
```diff
@@ -37,10 +37,12 @@ dependencies:
...
   - setuptools
   - six
+  - sphinx
   - trollimage
   - trollsift
   - xarray
   - scipy
   - pip:
     - git+https://github.com/pytroll/satpy.git@master
     - graphviz
```
wroberts4: @djhoese Do we need graphviz in pip and conda?

djhoese: iirc one is the C program (from conda) and one is the python library (pip) that is used in sphinx...I think.

build_environment.yml
```diff
@@ -16,7 +15,6 @@ dependencies:
   - fontconfig
   - freetype
   - gdal
-  - graphviz
```
djhoese: Can you check the jenkins logs, I'm pretty sure this provides the `dot` command which is needed for parts of the sphinx documentation.

djhoese: 🤦‍♂ Ah I see, it is in the other yaml file.

integration_tests/run.sh
```diff
@@ -1,37 +1,56 @@
...
 # Only ran by Jenkins if build was successful.
-# Save software bundle.
-rm -rf /tmp/polar2grid-*
-mkdir /tmp/polar2grid-$time
-cp -r "$WORKSPACE/$tarball_name" /tmp/polar2grid-$time
+# Remove old software bundles.
+rm -rf /tmp/${prefix}2grid-*
+mkdir /tmp/${prefix}2grid-${end}
+# Save software bundle and tarball.
+cp -r "$WORKSPACE/$swbundle_name" /tmp/${prefix}2grid-${end}
```
djhoese: If you are copying the tarball I'm not sure you need to copy the extracted directory. Thoughts?

integration_tests/run.sh
```diff
@@ -1,37 +1,56 @@
...
-set -exf
+set -ex
 export PATH="/usr/local/texlive/2019/bin/x86_64-linux":$PATH
 cd "$WORKSPACE"
 # Activate conda for bash.
 /data/users/davidh/miniconda3/bin/conda init bash
 # Restart the shell to enable conda.
 source ~/.bashrc
-conda env update -n jenkins_p2g_env -f build_environment.yml --prune
+conda env update -n jenkins_p2g_env -f build_environment.yml
```
djhoese: Since you aren't pruning anymore (which I think I told you not to do because of performance issues), doesn't the environment always include all of the "jenkins_environment.yml" packages? Another possibly dumb option would be have one "build" environment (only build_environment.yml is used) and one "test" environment for running behave and building documentation. Not sure it is worth it but since you are using `conda env update` maybe it wouldn't be that bad after the first creation.

build_environment.yml
wroberts4: It is required for sphinx, but only when building the docs which jenkins_environment.yml handles.

integration_tests/run.sh
wroberts4: I'm okay with that, I just saw that both of each version were in data/dist, and if they get unpacked anyways why not just put both the tarball and unpacked copies in data/dist.

integration_tests/run.sh
wroberts4: Pruning is bugged and doesn't actually remove any packages not listed in the yaml file. I like the idea of having two environments. The extra time invested in to making two environments is relatively minuscule and it makes an impact on the size of the tarball.

integration_tests/run.sh
djhoese: The tarballs won't always be unpacked. If we prove that they work there is a good chance Kathy and I can just take them and give them to people. Or Kathy might download them to another server and unpack them there. Overall, it just limits how much disk space is used.

integration_tests/run.sh
```diff
@@ -1,37 +1,58 @@
...
 
-time=`date +"%Y%m%d-%H%M%S"`
-tarball_name=polar2grid-swbundle-$time
-./create_conda_software_bundle.sh $WORKSPACE/$tarball_name
-export POLAR2GRID_HOME="$WORKSPACE/$tarball_name"
+# Handle release vs test naming.
+prefix="$(cut -d'-' -f1 <<<"$GIT_TAG_NAME")"
+end="`date +"%Y%m%d-%H%M%S"`"
+# If the tag is correct and a version was specified, make a version release.
+if [[ "$GIT_TAG_NAME" =~ [pg]2g-v[0-9]+\.[0-9]+\.[0-9]* ]]; then
```
djhoese: I think you need a `.` (period) just before the last asterisk otherwise only numbers are allowed after the version number, right? I guess I'm wondering if `1.2.3a0` is allowed.

integration_tests/run.sh
```diff
@@ -1,37 +1,58 @@
...
+    # Removes prefix from $GIT_TAG_NAME.
+    end=${GIT_TAG_NAME#"$prefix-"}
```
djhoese: Does this work with the quotes?

djhoese: It does, cool!

integration_tests/run.sh
```diff
@@ -1,37 +1,58 @@
...
+fi
+if [[ "$prefix" = "g2g" ]]; then
+    prefix=geo
+else
+    prefix=polar
+fi
+swbundle_name=${prefix}2grid-swbundle-${end}
+
+./create_conda_software_bundle.sh ${WORKSPACE}/${swbundle_name}
```
djhoese: *Iff* you have to make other changes, could you wrap this path in quotes. If either of the variables have spaces in them then they will be treated as separate arguments (99% sure).

integration_tests/run.sh
wroberts4: p2g-v1.2.3a0 is allowed! the tag name "p2g-v1.2.3a0" returns "polar2grid-swbundle-v1.2.3a0".

djhoese: Ok so it is a partial pattern match. Then shouldn't the last `*` be a `+`? Right now it will accept `p2g-v1.2.` I think.

integration_tests/run.sh
```diff
@@ -1,37 +1,58 @@
...
 # Clear out intermediate results and rebuild for HTML document.
 make clean
 # Needs to be second since Jenkins makes an html in workspace from the file generated by this command.
-make html
-cp -r "$WORKSPACE"/doc/build/html /tmp/polar2grid-$time
-chmod a+rx /tmp/polar2grid-$time
\ No newline at end of file
+make html POLAR2GRID_DOC=${prefix}
+cp -r "$WORKSPACE"/doc/build/html /tmp/${prefix}2grid-${end}
+chmod a+rx /tmp/${prefix}2grid-${end}
```
djhoese: I think you need `chmod -R a+rX`

integration_tests/run.sh
```diff
@@ -1,37 +1,58 @@
...
+# Only copy to data/dist if the tag was correct and a version was specified.
+if [[ "$GIT_TAG_NAME" =~ [pg]2g-v[0-9]+\.[0-9]+\.[0-9]* ]]; then
```
djhoese: This should be `+` on the end too.

djhoese: Need the regex change here too

PR #192, opened 2019-07-31
Update fornav templates to work with C++11

djhoese:
Linux on Travis is now using gcc 5.4 which is using C++11 by default. This is causing failures for fornavs compilation because there are cases where I pass an integer to `isnan` (a builtin function). In C99 this is a macro and returned False for integers, in C++11 this is a real function that only has signatures for floats/doubles. This is my simple fix.

@rayg-ssec It'd be great if you could sanity check this given my limited C++ knowledge.

### DIFF
--- a/polar2grid/remap/_fornav_templates.cpp
+++ b/polar2grid/remap/_fornav_templates.cpp
@@ -4,6 +4,18 @@
 #include "numpy/arrayobject.h"
 #include "_fornav_templates.h"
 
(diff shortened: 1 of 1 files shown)


djhoese:
I should probably inline the functions although I would expect a compiler to figure out the functions could be inlined with how simple they are.

djhoese:
Apparently I already fixed this in the pyresample code. Last commit copies pyresample's logic which is equivalent to what I was doing here.

PR #201, opened 2019-08-14
Fix rescaling configuration for some active fires variables

djhoese:
Pointed out in our meeting today:

1. `confidence_pct` being from 0-100% results in `0` values (the most common value) being yellow which is different than `confidence_cat` which is transparent for non-fire pixels.
2. Fire `power` was not being scaled properly because I used the `reader` parameter in the rescaling configuration. The Satpy reader wrapping code was not assigning the `reader` in product metadata so this information was never getting to rescaling.

This PR should fix both of these cases.

### DIFF
--- a/polar2grid/core/rescale_configs/rescale.ini
+++ b/polar2grid/core/rescale_configs/rescale.ini
@@ -594,7 +594,7 @@ alpha=True
 product_name=confidence_pct
 method=palettize
 colormap=ylorrd
-min_in=0
+min_in=1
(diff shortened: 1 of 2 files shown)


djhoese:
Also, waiting on https://github.com/pytroll/satpy/pull/877 to be merged before making the next P2G alpha tarball since it includes fixes for the T4/T13 bands so that P2G can know how to rescale them based on units/standard_name.

PR #202, opened 2019-08-15
Improve Compare Scripts

wroberts4:
I added margin_of_error to polar2grid.compare which allows for a percent of total pixels to be wrong. For example if margin_of_error=1, then 1 out of every 100 pixels can be wrong (whereas each pixel's correctness is determined by atol).

I also made p2g_compare_netcdf.sh and p2g_compare_geotiff.sh use polar2grid.compare for the compare logic.


djhoese:
I talked to @kathys and we think it would be best if the margin stayed 0 by default. After you make the changes I recommended (calling the python module directly) you could allow additional command line arguments to be passed to the python script (see polar2grid.sh for `$@` usage example).

integration_tests/run.sh
```diff
@@ -26,17 +20,37 @@ else
...
+# Restart the shell to enable conda.
+source ~/.bashrc
+
 # Documentation environment also has behave, while the build environment does not.
+conda env update -n jenkins_p2g_docs -f "$WORKSPACE/build_environment.yml"
 conda env update -n jenkins_p2g_docs -f "$WORKSPACE/jenkins_environment.yml"
 conda activate jenkins_p2g_docs
+pip install "$WORKSPACE"
+
+if [[ "$GIT_TAG_NAME" =~ [pg]2g-skip-tests ]]; then
```
djhoese: We don't want to make these types of decisions based on the tag, but rather on the commit message. I don't remember how jenkins makes this available `GIT_COMMIT_MSG` maybe? It's in the setting.

We don't want to make a new tag every time we want to run tests in a special way. Also there would be no way to stop the tests from running in master if it was dependent on tags. The way Travis and other CIs typically do this is something like:

```
git commit -m "My normal message [skip-ci]
```

We could use this idea to detect `[skip-tests]` or `[p2g-skip-tests]` or something.

swbundle/p2g_compare_geotiff.sh
```diff
@@ -59,34 +59,13 @@ BAD_COUNT=0
...
-equal_pixels = len(numpy.nonzero( work_data == valid_data )[0])
-if equal_pixels != total_pixels:
-    print ("FAIL: %d pixels out of %d pixels are different" % (total_pixels-equal_pixels,total_pixels))
-    sys.exit(2)
-print ("SUCCESS: %d pixels out of %d pixels are different" % (total_pixels-equal_pixels,total_pixels))
-EOF
-    [ $? -eq 0 ] || BAD_COUNT=$(($BAD_COUNT + 1))
+    echo "INFO: Comparing $WFILE to known valid file"
+    python -m polar2grid.compare geotiff "$VFILE" "$WFILE" --atol 0. -vv
+[ $? -eq 0 ] || BAD_COUNT=$(($BAD_COUNT + 1))
```
djhoese: Should this be indented (visually) for clarity?

swbundle/p2g_compare_geotiff.sh
```diff
@@ -59,34 +59,13 @@ BAD_COUNT=0
...
-total_pixels = work_data.shape[0] * work_data.shape[1]
-equal_pixels = len(numpy.nonzero( work_data == valid_data )[0])
-if equal_pixels != total_pixels:
-    print ("FAIL: %d pixels out of %d pixels are different" % (total_pixels-equal_pixels,total_pixels))
-    sys.exit(2)
-print ("SUCCESS: %d pixels out of %d pixels are different" % (total_pixels-equal_pixels,total_pixels))
-EOF
-    [ $? -eq 0 ] || BAD_COUNT=$(($BAD_COUNT + 1))
+    echo "INFO: Comparing $WFILE to known valid file"
+    python -m polar2grid.compare geotiff "$VFILE" "$WFILE" --atol 0. -vv
```
djhoese: I think we talked about maybe letting users pass additional arguments by passing `"$@"` here, right? Then users could add more `-v` if they wanted or change `--atol` themselves. I think `atol` should probably default to `0` in the python script.

For the argument passing you could look at using the `shift 1` command when the arguments are read above (the `VERIFY_BASE` and `WORK_DIR` part). This would make `$@` be the arguments we care about. Thoughts?

swbundle/p2g_compare_geotiff.sh
wroberts4: I believe so. But the netcdf compare script didn't use an indent, and I'm not sure what the purpose of line 68 was. I decided to be safe and not indent it incase it had some kind of shell script meaning behind it.

djhoese: Indentation doesn't have any meaning in bash (at least 99% of the time). I'd say indent it for clarity. I'm assuming that github is showing the code to me correctly and that is all part of the same block of code.

integration_tests/README.rst
```diff
@@ -3,13 +3,16 @@ How to run the tests in own environment
 
 Navigate to your Polar2Grid directory, then run::
 
+    conda env update -n jenkins_p2g_docs -f build_environment.yml
+    conda env update -n jenkins_p2g_docs -f jenkins_environment.yml
```
djhoese: Just curious, is it possible to list multiple YAML files to a single `conda env update` call?

integration_tests/README.rst
```diff
@@ -3,13 +3,16 @@ How to run the tests in own environment
...
+    conda activate jenkins_p2g_docs
+    pip install .
+    conda env update -n jenkins_p2g_swbundle -f build_environment.yml
+    conda activate jenkins_p2g_swbundle
+    ./create_conda_software_bundle.sh polar2grid-YYYYMMDD-HHmmSS                                    # Creates a folder with the necessary scripts to run the tests
```
djhoese: You _could_ use ``datetime +%Y%m%d-%H%M%S`` (need the single back ticks) to place the date in the filename. Not necessary to change, but FYI. You may want to change the `MM` and `DD` in the above since they are the wrong case.

wroberts4: Interesting, good to know. I have seen MM used for months also, but I will change this and keep it in mind for the future (https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html)!

PR #208, opened 2019-08-21
Fix MERSI-2 L1B generation of true_color RGB

djhoese:
@kathys pointed out that in the 2.3.0a2 tarball the mersi2_l1b reader doesn't generate the true_color composite. For some reason it delays the generation (sunz angle depending on 1000m angles) but never regenerates them after resampling is complete.

@kathys Feel free to add information here (test files, log file locations, etc) if you haven't already started an issue.

### DIFF
--- a/build_environment.yml
+++ b/build_environment.yml
@@ -39,4 +39,4 @@ dependencies:
   - xarray
   - scipy
   - pip:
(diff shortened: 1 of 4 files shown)


kathys:
I placed the data used and the log file that resulted at this location:

bumi:/data/users/kathys/test_data/fy3d_mersi/day

PR #230, opened 2019-10-28
Fix viirs_l1b standard_name Not Found

wroberts4:
viirs_l1b.py was looking for standard_name where it should not have been. This is a work around that does not generate composites when composites should not be made.


djhoese:
This is another case of "complex solution for a simple problem because we aren't completely dependent on Satpy yet". Thanks @wroberts4.

PR #231, opened 2019-11-01
Fix race condition when deleting multiple gridded scenes

djhoese:
This was an issue that @kathys identified but had trouble nailing down the exact cause and reproducing it. She sent me a log for one such case and I have an idea for what's going on. Bottom line this should fix this issue.

## Details

Doing `del gridded_scene` in python (see code changes for this PR) is meant to delete the current gridded scene and let python's garbage collector remove things when it wants. As part of deleting the gridded scene, the flat binary files underneath should also be deleted (ex. `true_color.dat`). However, it seems that in some cases this is taking too long and causing issues. Mainly this happens when resampling to multiple grids. If we process grid 1, delete the gridded_scene, move on to process grid 2, we end up creating/using the same flat binary file as the first grid's processing. This can result in P2G deleting the binary file we are currently using, causing issues later in the processing.

## Alternate Solution

I could have also made the filename more unique but I think this should work and is a simpler solution.

### DIFF
--- a/polar2grid/compositors/rgb.py
+++ b/polar2grid/compositors/rgb.py
@@ -237,7 +237,9 @@ class RGBCompositor(roles.CompositorRole):
         if fill_value is None:
             fill_value = gridded_scene[self.composite_products[0]]["fill_value"]
 
-        fn = self.composite_name + ".dat"
+        base_product = gridded_scene[self.composite_products[0]]
+        grid_name = base_product['grid_definition']['grid_name']
+        fn = "grid_{}_{}.dat".format(grid_name, self.composite_name)
 
         try:
             comp_data = self.joined_array(gridded_scene, self.composite_products)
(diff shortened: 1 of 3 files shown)


djhoese:
For the record, I'll explain what this latest commit does in English: There were two major bugs in Polar2Grid that apparently weren't a problem until newer versions of python where garbage collection priorities have changed.

1. All composites were using the same `.dat` filename on disk. This meant that there was a race condition where a `true_color.dat` could be written for one grid, then during the processing of the next grid it could be deleted at any time depending on python's garbage collection and the `__del__` methods used in python to "magically" delete the files when the associated product was deleted from the code. If the `true_color` product from the previous grid processing was deleted/garbage collected after the current iteration had already created its composite then that current product didn't have its data any more.

2. When creating a remapped "GriddedProduct" and using the metadata from the "SwathProduct" that was just resampled, I purposely removed some metadata like `swath_definition`, `swath_cols`, and `swath_rows` since they no longer applied to this gridded product. However, I wasn't actually using this new "filtered" metadata and was mistakenly using the original swath metadata. This was causing a SwathDefinition to exist and be copied when it should not have been. This bug is what was causing the `m_lons.dat` errors we were sometimes seeing.

PR #234, opened 2019-11-13
Add email information to Jenkins and improve style

wroberts4:
I added a variables.txt file in integration_tests that Jenkins email extension can read and add information to emails. The compressed build log is also attached to the email. An example:

####################################################################
Here is information from the most recent polar2grid/geo2grid tests.

General:
&nbsp;&nbsp;&nbsp;&nbsp;Start time: 2019-11-11 23:29:46
&nbsp;&nbsp;&nbsp;&nbsp;Finish time: 2019-11-12 01:47:49
&nbsp;&nbsp;&nbsp;&nbsp;Author: wroberts4@wisc.edu
&nbsp;&nbsp;&nbsp;&nbsp;Commit hash built: abcde
&nbsp;&nbsp;&nbsp;&nbsp;Commit message: This is a test [p2g]
&nbsp;&nbsp;&nbsp;&nbsp;Tag name:

Polar2grid:
&nbsp;&nbsp;&nbsp;&nbsp;Tests: SUCCESSFUL
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1.1 ACSPO: passed in 14 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1.2 ACSPO: passed in 7 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2.1 AMSR2_L1B: passed in 30 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;3.1 AVHRR: passed in 28 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;4.1 MODIS: passed in 56 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;5.1 VIIRS: passed in 106 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;6.1 VIIRS_L1B: passed in 166 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;7.1 VIIRS_SDR: passed in 320 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;7.2 VIIRS_SDR: passed in 67 seconds
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;7.3 VIIRS_SDR: passed in 29 seconds
&nbsp;&nbsp;&nbsp;&nbsp;Documentation: SUCCESSFUL
&nbsp;&nbsp;&nbsp;&nbsp;Package name: polar2grid-20191111-232946
&nbsp;&nbsp;&nbsp;&nbsp;Package sent to "bumi:/tmp": TRUE

Geo2grid:
&nbsp;&nbsp;&nbsp;&nbsp;Tests: SKIPPED
&nbsp;&nbsp;&nbsp;&nbsp;Documentation: SKIPPED
&nbsp;&nbsp;&nbsp;&nbsp;Package name:
&nbsp;&nbsp;&nbsp;&nbsp;Package sent to "bumi:/tmp": FALSE
#######################################################

integration_tests/run.sh was reorganized into functions in the hope of making the code more organized and easier to read. I also changed some double quotes to single quotes out of spite...


integration_tests/features/steps/compare_images.py
```diff
@@ -58,4 +58,3 @@ def step_impl(context, output):
         assert exit_status == 0, "Files did not match with the correct output"
     finally:
         os.chdir(orig_dir)
-        shutil.rmtree(context.temp_dir)
```
wroberts4: This wasn't always deleting files (ie when builds are interrupted, etc). I moved the removal of tmp files to Jenkins. Now tmp files are deleted before running run.sh and hence Jenkins will not run out of tmp space from tmp files, and we can debug easier by seeing the output files (since we don't delete them after the r=tests are ran).

djhoese: Why are builds interrupted? Manually I suppose? And/or does that include Jenkins detecting that a new build is being started so it can cancel currently running ones?

Can we retain the removal in the behavior tests so that they can be run outside of/without Jenkins?

integration_tests/run.sh
```diff
@@ -28,61 +28,194 @@
...
+
+        # Replace FAILED with SUCCESSFUL.
+        save_vars "${prefix:0:1}2g_tests=SUCCESSFUL"
+    )
+}
+
+create_documentation()
+{
+    set +e
+    (
```
djhoese: Could you explain what you're trying to do with this `set +e` followed by the parentheses?

wroberts4: Essentially a "try" clause. I want it to break out of that segment if an error occurs without breaking out of the entire program. This allows finish_time to be made, AND allow documentation to be tested even if the tests failed.

integration_tests/features/steps/compare_images.py
wroberts4: Sure, I'll have to test why not all tmp files are being removed. There's usually a couple left over.

integration_tests/run.sh
wroberts4: Would you rather have the documentation fail if tests fail? I think there should be a try clause around the main part of the program so finish_time is saved regardless of any error. I can remove the other try clauses of you think the documentation should fail if tests fail.

djhoese: Wow, really cool! I think the way you have it is good then. We'll change it in the future if it becomes a problem. We don't want documentation to be generated and **uploaded** if tests fail, but generating it alone should be fine.

integration_tests/run.sh
```diff
@@ -28,61 +28,221 @@
...
+            end = ''
+        print("\t\t{0}: {1} in {2} seconds".format(name, test['status'], round(duration)), end=end)
+EOF
+    set -x
+    rm "$json_file"
+}
+
+run_tests()
+{
+    try
```
djhoese: What is the `try` function if you just reset the `set -e` call in the subshell? What is the `try` function doing that isn't equivalent to calling `set +e`?

integration_tests/run.sh
```diff
@@ -28,61 +28,221 @@
...
+            if [[ "$commit_message" =~ (^|.[[:space:]])"["([pg]2g-)?skip-tests"]"$ ]]; then
+                # Replace FAILED with SKIPPED.
+                save_vars "${prefix:0:1}2g_tests=SKIPPED"
+            else
+                run_tests $prefix
+                update_exit_status ${exit_status}
+            fi
+            create_documentation $prefix $package_name
+            # Only publishes if both tests and documentation passed.
+            if [[ ${exit_status} -eq 0 ]]; then
```
djhoese: Is this actually checking the documentation exit status at all?

integration_tests/run.sh
```diff
@@ -28,61 +28,221 @@
...
+            fi
+            exit ${exit_status}
+        )
+        update_exit_status ${exit_status}
+    done
+    exit ${exit_status}
+)
+update_exit_status ${exit_status}
+# Allows errors to happen in save_vars.
+set -e
```
djhoese: `set -e` does the opposite of allowing errors in `save_vars`

.travis.yml
```diff
@@ -21,6 +21,7 @@ matrix:
 install:
   - git clone --depth 1 git://github.com/astropy/ci-helpers.git
   - source ci-helpers/travis/setup_conda.sh
+  - conda install -c conda-forge zarr pyspectral
```
djhoese: pyspectral is already installed via the environment variables at the top of this file. Zarr should be added to the conda dependencies.

wroberts4: Done!

integration_tests/run.sh
wroberts4: It is now. If documentation fails, then the code will exit the subshell before trying to publish.

integration_tests/run.sh
wroberts4: I agree that using too many subshells is confusing. I tried to limit their use and instead use OR and AND statements where needed to simplify the logic.

Issue #236, opened 2019-11-15
Issues with polar2grid and AVHRR clavrx products..


spruceboy:
I am trying to make geotifs and scmi tiles from AVHRR clavrx products, and am running into problems.  Maybe I am doing it wrong? 

With the lastest version of polar2grid i have (2.3.0a2), it generates an error message.
```
[jecable@roundrect tst]$ ./polar2grid-swbundle-2.3.0a2/bin/polar2grid.sh  clavrx gtiff  -g 203 -p cld_temp_acha -f ./in/clavrx_hrpt_noaa19_20191031_1910_55298.l1b.level2.hdf
INFO     : Initializing reader...
ERROR    : clavrx frontend failed to load and sort data files (see log for details)
```
I have attached the log file. 

[clavrx2gtiff_fail.log](https://github.com/ssec/polar2grid/files/3849498/clavrx2gtiff_fail.log)

Using an older version of polar2grid (2.2.1) it seems to just run forever:

```
./polar2grid_v_2_2_1/bin/polar2grid.sh  clavrx gtiff  -g 203 -p cld_temp_acha -f ./in/clavrx_hrpt_noaa19_20191031_1910_55298.l1b.level2.hdf
INFO     : Initializing reader...
INFO     : Initializing remapping...
INFO     : Initializing backend...
INFO     : Initializing compositor objects...
INFO     : Extracting swaths from data files available...
/home/jecable/tst/polar2grid_v_2_2_1/common/ShellB3/lib/python2.7/site-packages/pyresample/geometry.py:86: UserWarning: All geometry objects expect longitudes in the [-180:+180[ range. We will now automatically wrap your longitudes into [-180:+180[, and continue. To avoid this warning next time, use routine utils.wrap_longitudes().
  'To avoid this warning next time, use routine utils.wrap_longitudes().')
INFO     : Remapping to grid 203
INFO     : Grid information:
        grid_name: 203
        proj4_definition: +proj=stere +a=6371200 +b=6371200 +lat_0=90 +lat_ts=60.0 +lon_0=-150 +units=m +no_defs
        cell_width: 1000.0
        cell_height: -1000.0
        width: 8384
        height: 7239
        origin_x: -4952960.85604
        origin_y: 381022.610378
WARNING  : Data has less than 2 rows per scan, this is not optimal for the EWA resampling algorithm. All rows will be used as one scan
..
```

I am probibly doing something wrong, if so please let me know what i am missing. 

The file I am using for testing is here:
[...]

PR #238, opened 2019-11-16
Fix EWA fornav resampling hanging when geolocation includes NaNs

djhoese:
This was first noticed by @spruceboy when processing AVHRR data from clavrx. This PR closes the related issue #236 

For some reason the data files are being produced with the first 8 or so rows of geolocation (longitude and latitude) being all missing. In Polar2Grid these missing values are represented as NaNs. The calculations performed in fornav with the geolocation (actually the columns and rows from the ll2cr step) end up with a lot of the algorithm parameters being NaN. This causes EWA to either not fall in to or fall in to `if` satements that normally wouldn't be done causing processing to take an extremely long time (~1-5 rows per second compared to 100s or 1000s of rows per second).

This PR tries to address this by resetting the NaN parameters to the maximum distance parameter to limit how many pixels are processed when geolocation is NaN.


djhoese:
Note, this PR also includes the changes from #237

PR #241, opened 2019-12-02
Fix ll2cr code compatibility with pyproj 2.4.2

djhoese:
Pyproj 2.4.2 changes the invalid transformation value from 1e30 to `inf`. In most cases checking for `>= 1e30` is sufficient, but this PR fixes the one case where I don't do that.


djhoese:
This is failing because zarr needs to be installed from conda. I believe @wroberts4 has fixed this in #234 so I'm going to merge this since the linux environment is passing.

PR #251, opened 2019-12-18
Update documentation for Polar2Grid 2.3

djhoese:


djhoese:
@kathys Our unit tests are failing because the example images you reference for the documentation are not on the FTP location. Could you add those?

PR #253, opened 2020-01-08
Add Scripts to Environment

wroberts4:
I added polar2grid and geo2grid scripts to the environment. When users run `pip install polar2grid`, they will then be able to run `polar2grid.sh -h` directly from the environment and run it as normal. This can also be done with all other shell scripts in `swbundle` with the exception of `env.sh` and `polar2grid_env.sh`.

This was done by adding these shell scripts to the `scripts` key in setup.py. When running the scripts from the environment, they do not call `env.sh`. In it's place is `# __SWBUNDLE_ENVIRONMENT_INJECTION__` which is replace with `source $POLAR2GRID_HOME/bin/env.sh` by using sed replace when the swbundle is made.

Additionally, I added a SWBUNDLE_OR_ENVIRONMENT variable to integration_tests/run.sh which determines wether or not to run tests using the swbundle or strictly the environment (0 being swbundle, 1 being the environment).

Finally, I cleaned up some files, code structure, and documentation. The README for integration_tests now gives information on how to add more tests.


create_conda_software_bundle.sh
```diff
@@ -112,14 +112,13 @@ cd $SB_NAME/bin
...
-cp -r $BASE_P2G_DIR/etc/* $SB_NAME/etc/satpy/ || oops "Couldn't copy configuration 'etc' directory"
-
-conda init bash
-# Restart the shell to enable conda.
-source ~/.bashrc
-conda deactivate
+# Inject environment code into swbundle only.
+for file in `echo *.sh`; do
+    cp "$file" tmp.sh
+    sed "s/# __SWBUNDLE_ENVIRONMENT_INJECTION__/source \$POLAR2GRID_HOME\/bin\/env.sh/g" tmp.sh > "$file"
```
djhoese: Isn't there a `-i` flag for inplace modification?

integration_tests/features/polar2grid.feature
```diff
@@ -21,12 +21,12 @@ Feature: Test polar2grid output images
     | polar2grid.sh avhrr gtiff -vv -f | avhrr/input/test1 | avhrr/output/test1 |
 
   Examples: MODIS
-    | command                                                                                                                                             | source            | output             |
-    | polar2grid.sh crefl gtiff -vv --true-color --false-color --fornav-D 10 --grid-configs ${POLAR2GRID_HOME}/grid_configs/grid_example.conf -g miami -f | modis/input/test1 | modis/output/test1 |
+    | command                                                                                                                                      | source            | output             |
+    | polar2grid.sh crefl gtiff -vv --true-color --false-color --fornav-D 10 --grid-configs ${DATAPATH}/grid_configs/grid_example.conf -g miami -f | modis/input/test1 | modis/output/test1 |
```
djhoese: DATAPATH seems like an odd name for this. I would assume DATAPATH is where the test data and expected data are located. This seems like it is where the tarball/bundle is, right?

integration_tests/features/steps/compare_images.py
```diff
@@ -11,7 +11,7 @@ def step_impl(context, source):
     new_source = ""
 
     for f in source.split(" "):
-        f = os.path.join(context.data_path, f)
+        f = os.path.join(os.environ['DATAPATH'], f)
```
djhoese: Why environment variable over context attribute?

polar2grid/glue.py
```diff
@@ -47,6 +47,8 @@
 except ImportError:
     CRS = None
 
+os.environ["PPP_CONFIG_DIR"] = os.path.join(sys.prefix, 'etc', 'satpy')
```
djhoese: You may want to use `os.path.dirname(__file__)` or something similar. Maybe pkg_resources has a way to get the path of the `etc/satpy` directory?

What do you think about changing this to `etc/polar2grid` too? I know this would require find/replacing it everywhere that it is used but now that this will be installed this way any time polar2grid is installed, it would be best to avoid satpy's configuration (right now this doesn't because of how satpy is set up, but we should be careful for future changes).

integration_tests/features/polar2grid.feature
wroberts4: DATAPATH is where the test data is stored, yes. the config files are with the rest of the test data (/data/test_data).

djhoese: Ooohhh ok. Sounds good.

integration_tests/features/steps/compare_images.py
wroberts4: I need to pass it into the scripts being called, but I am going to use just a context attribute and then pass it in when the commands are called like so `'DATAPATH={} ...'.format(context.data_path)`

create_conda_software_bundle.sh
wroberts4: Sed replace in-place is not very universal. There are different requirements between shells and versions, so this is much more maintainable to just explicitly make a temporary file. for example when I do `sed -i "s/# __SWBUNDLE_ENVIRONMENT_INJECTION__/source \$POLAR2GRID_HOME\/bin\/env.sh/g" tmp.sh` I get the error `sed: 1: "tmp.sh": undefined label 'mp.sh'`

djhoese: Ok.

polar2grid/glue.py
wroberts4: I have no preference for the naming. I feel like naming is satpy makes sense since it's data for satpy, but I'm fine with polar2grid. The problem with os.path.dirname(__file__) is that the file will be in site_packages, while the config directory is at the base level. I guess I could manually have it traverse back some levels, but that feels like an error waiting to happen.

djhoese: Agreed on the error waiting to happen. See if the `pkg_resources` package can be used to tell you where the `etc` directory is. `pkg_resources` should know where `setup.py` installed things.

Altlhough these configs get used by Satpy, we are placing them in `<prefix>/etc/satpy` which is a "global" location that may be used by satpy. Only Satpy (and optionally satpy plugins in the future) should be modifying things in `etc/satpy`. Note that Satpy doesn't currently use `etc/satpy` but that may change in the future. This is my thinking at least.

wroberts4:
Note: We will need to change "rm -rf /tmp/tmp*" to "rm -rf /tmp/p2g_tests_*" in the execute shell part of Jenkins config when this is merged

PR #266, opened 2020-03-10
Clear LD_LIBRARY_PATH in conda-based distributions

djhoese:
A recent mailing list post had issues with finding the NetCDF library. This made me think that they had `LD_LIBRARY_PATH` set which shouldn't have been a problem because the `env.sh` script is supposed to remove that. However, after looking I noticed that it only does this for the legacy ShellB3 builds of Polar2Grid. This PR fixes this by doing it for conda-based distributions too.


djhoese:
No, but I'll address that in a future PR. These if statements and names were needed when P2G was still using ShellB3 but G2G was using conda-pack.

Issue #227, opened 2019-10-24
Document image types


graemely:
For V1.1, it would be helpful to include in the documentation a list of the types of images that can be generated for each instrument, and a brief explanation of each type of image (especially composite RGBs).

djhoese:
Do you have an example of what you're looking for? I think some basic information is already provided in the documentation (both the PDF and the website):

What reader/writers are available and what their filenames look like or what is produced (formatting is messed up for some reason on the tables): https://www.ssec.wisc.edu/software/geo2grid/getting_started.html#reader-writer-combinations

And each reader section has a list of the products that can be requested: https://www.ssec.wisc.edu/software/geo2grid/readers/abi_l1b.html

graemely:
The list of products is what I was looking for, but it appears to be incomplete. There is a much larger list of products that is generated with `--list-products`, running v1.0.

djhoese:
:thinking: 

That may be a bug or may be intended. Their are the composites that we "support" meaning we've looked at them and they look the way the should, and there are products that Satpy has builtin or that the G2G user created custom which we can't guarantee. In P2G we limit what is shown with `--list-products` even though the user can ask for a lot more (at least from Satpy's readers). In G2G, I think we decided early on that the user should see their custom composite name listed when they do `--list-products`.

I'm not sure there is an easy way around that besides mentioning in the documentation that other RGBs may be shown or by removing all the Satpy builtins (not sure that is possible in Satpy right now). Note that *some* composites may be "intermediate" composites that aren't meant to be listed and newer versions of G2G may properly remove them from the list.

graemely:
Is there any way to list only the verified products and the user-created products? I think it would be preferable not to list products we have not verified.

Also is there any reason not to generate the full set of verified products by default?

djhoese:
> Is there any way to list only the verified products and the user-created products?

Right now, no. Satpy (the code that is loading the composite recipes) treats the user created recipes similar to the builtin ones. So right now it isn't possible in Satpy to distinguish.

> Also is there any reason not to generate the full set of verified products by default?

You mean generate all the possible RGBs if the user doesn't specify `-p`? No. We just decided that most people aren't going to need/want them just to draw the line somewhere.

djhoese:
[...]

PR #301, opened 2020-12-05
Add initial "Complex Composites" documentation

djhoese:
This adds a new page to the HTML website about "Complex Composites". It includes the general/common corrections and modifications that people might see and then describes a couple of the composites that use these. Right now this only includes the ABI True Color.

Related #227 

@graemely @kathys If you click on "Files changed" above you should be able to comment on specific lines of my changes. Start with the page that says "Complex Composites" and then move to the "geo2grid_composites.rst" file. Let me know if you have any issues with this or other ideas.


djhoese:
I'm still not sure how I feel about this showing up outside of the reader documentation, but agree with what @kathys said on slack that it may be too much detail to throw in the "abi_l1b" reader documentation; even if I only have it show up on the website.

doc/source/complex_composites/index.rst
```diff
@@ -0,0 +1,100 @@
...
+for the amount of sun reflecting off of the atmosphere. While this is a very
+common change when working with numeric reflectance data, some RGB recipes
+may not include it.
+
+.. _psp_rayleigh_correction:
+
+Rayleigh Scattering Correction - Pyspectral
+-------------------------------------------
+
+Due to atmospheric molecules, some visible channels are scattered obscuring
```
kathys, line 25: Due to the size of molecules that make up our atmosphere, some visible channel light is preferentially scattered more than others, especially at larger viewing angles.

doc/source/complex_composites.rst
```diff
@@ -0,0 +1,120 @@
...
+combining them we can produce an array of colors. This differs from a
+basic grayscale or "luminance" image where the single "L" channel
+determines how white or black a pixel is. For more information on the
+RGB color model see
+`this Wikipedia article <https://en.wikipedia.org/wiki/RGB_color_model>`_.
+
+You may also see RGBA images where the A stands for "Alpha". This Alpha
+channel will determine the opaque or transparency of an image. This is
+typically used in |project| to make invalid or missing data values
+transparent (completely opaque or completely transparent).
```
graemely, line 27: @djhoese rather than describing the RGB color model how about something like:
Satellite imagers can simultaneously observe the Earth in multiple spectral channels, while the human eye is sensitive to only the visible channels. By mapping data from imager channels to the visible red, green and blue channels in different ways, we can produce “RGB” images that show the Earth as a human would see it from space (“true color”), or that emphasize certain features that can be detected using combinations of different channels (“false color”).

doc/source/complex_composites.rst
```diff
@@ -0,0 +1,120 @@
...
+
+.. _sunz_correction:
+
+Solar Zenith Angle Modification
+-------------------------------
+
+Visible channel data is not truly a "reflectance" until it has been adjusted
+for the amount of sun reflecting off of the atmosphere. While this is a very
```
graemely, line 35: Isn't it the incident radiation that needs to be accounted for?

doc/source/complex_composites.rst
```diff
@@ -0,0 +1,120 @@
...
+common change when working with numeric reflectance data, some RGB recipes
+may not include it.
```
graemely, line 37: Does this mean some RGB recipes used by G2G, or other software?

djhoese: Other software. For example the CIMSS True Color recipe does not apply this adjustment.

doc/source/complex_composites.rst
```diff
@@ -0,0 +1,120 @@
...
+pixels), then upsample the result again. By taking the ratio of the original
+high resolution and this averaged version, we can produce a ratio similar
+to that in the above ratio sharpening technique.
+
+.. _nonlinear_true_color_scaling:
+
+Non-linear True Color Scaling
+-----------------------------
+
+As a final "enhancement" or scaling method, Geo2Grid uses a series of linear
```
graemely, line 93: Can we add something to make it clear the intensity is being scaled, rather than something spatial. For example "As a final step, Geo2Grid scales the observed or calculated values to a standard range of color intensities, using a series of..."

doc/source/complex_composites.rst
djhoese: That sounds good. I'll try to add something like that. I actually meant to mention satellite imagery here, but forgot after bouncing between sections. I'd still like to talk about RGB images (the color model) a small bit, but I'll see if I can make it fit better.

doc/source/readers/abi_l1b.rst
```diff
@@ -27,3 +27,85 @@ Examples:
...
+    Product Explanation
+    -------------------
+
+    True Color
+    ^^^^^^^^^^
+
+    The ABI L1b "true_color" composite produced by Geo2Grid provides a view
+    of the Earth as the human eye would see it; or as close as we can come
+    to with satellite data and the channels we have available. This means
+    things like trees and grass are green, water is blue, desserts are
```
graemely, line 42: "deserts"

doc/source/complex_composites.rst
djhoese: I don't know. Someone please correct me (@kathys?).

doc/source/readers/abi_l1b.rst
djhoese: Ha, apparently I had something on my mind.

doc/source/complex_composites.rst
djhoese: To clarify, as far as I know there are no default RGBs in G2G that don't apply this adjustment to reflectance channels.

graemely: If I understand it correctly, how about something like this for the whole section:
Reflectance is defined as the reflected radiation as a fraction of the incident radiation. To calculate reflectance, the solar zenith angle is needed, in addition to the radiance measured by the sensor. Some of the RGB recipes applied by Geo2Grid use reflectance (with the solar zenith angle modification), and others use radiance directly.

djhoese: Oh well now by using "radiance" in that last part you've opened up a can of worms. Geo2Grid applies various calibration steps to convert the digital numbers in the file to a "normalized radiance" (at least I think that's what @kathys calls it). This is different than the radiance that you can also convert these digital numbers to. This issue has resulted in a huge discussion in the pytroll community, with Kathy, and with the GOES-R Imagery team. See https://github.com/pytroll/satpy/issues/536# for some of the pytroll discussion.

graemely: Happily, you can scratch the last sentence because I wrote it before I read your comment and it sounds like it is not needed anyway.

djhoese: Just realized the other important part here is that the individual channel images (C01-C06) don't have this applied.

djhoese:
@graemely I think I've updated everything you've commented on. If it looks "good enough" I'll merge it and update the website.

djhoese:
Merging now after discussion. There may be more updates in the future to clear things up or get them reorganized, but I think this is a good start.

Issue #269, opened 2020-03-25
Port legacy Polar2Grid readers to Satpy


djhoese:
These readers exist in legacy Polar2Grid but are not available in Satpy yet:

- [x] MIRS
- [x] VIIRS CREFL
- [x] MODIS CREFL

I would consider migrating the crefls as unnecessary since Satpy has a python implementation of that software (which could use some optimization).

djhoese:
@kathys What about the DR-RTV (Dual Regression - Retrieval) algorithm reader? Is that something we still support?

kathys:
No, we do not have to support the Dual Regression retrievals anymore.

djhoese:
@katherinekolman In addition to your SIFT work, could you work on migrating the MIRS reader to Satpy? All of the existing code is [here](https://github.com/ssec/polar2grid/tree/master/polar2grid/mirs). I see there is some test data on `bumi:/data/test_data/old_polar2grid_data/HI_MIRS` but I'm not sure if that is still valid. @kathys do you have a recent location where Kat can get new test files?

kathys:
I can get some to you @katherinekolman.  I will try and get it to you before the end of the day tomorrow.

djhoese:
MODIS and VIIRS CREFL should now be used by creating the `true_color` and `false_color` composites from the `modis_l1b` and `viirs_sdr` readers.

Issue #151, opened 2018-09-27
Some data issues with SFR mirs


spruceboy:
We are noticing some issues with MIRS SFR data with n19 and metopb, and are wondering how the "extra" missing data values are handled. 

The metadata for those files says:

```
                :missing_value = -999 ;
                :notretrievedproduct_value = -888 ;
                :noretrieval_value = -99 ;
```
 Some Examples of what this looks like:
![sfr_amsu_180927_1100](https://user-images.githubusercontent.com/243433/46179428-9f031700-c267-11e8-8ded-f6a161c37947.png)
![sfr_amsu_180927_1241](https://user-images.githubusercontent.com/243433/46179429-9f031700-c267-11e8-9624-3a88b501c49d.png)

Looking at the raw netcdf files with gdal, it looks to me like the regions that have the bad values show up as nodata, making wonder if gdal is masking off the other two nodata values? 

This could be a mistake on our end though.

djhoese:
What a strange set of fill values. iirc the MIRS reader/frontend only looks at the `missing_value`. If those other values are the problems then we'll have to add some extra logic.

spruceboy:
The ATMS data looks much more like what we expect: 
![sfr_amsu_180927_1355](https://user-images.githubusercontent.com/243433/46179818-8e53a080-c269-11e8-9aed-e6800f82bac5.png)

We (obviously) could be doing something wrong.   The NWS appears to be a lot of interest in the SFR (and the precipitation ) MIRS products, so if this is something I am doing wrong, let me know.

spruceboy:
Thanks for the awesome tool, btw!

djhoese:
Hello, 3.5 years later!

I'm going over old P2G issues and came across this one, but have no memory of if it was resolved or not (possibly over email). @spruceboy Not sure if you still get github notifications, but here we go...

There have been a lot of MiRS processing updates (not P2G) since this original issue. The little bit that I can remember is that this is/was likely a bug in the MiRS processing software where it should have marked these as fills. Even the new satpy-based readers for MiRS files do not check these other fill values. If anyone (@spruceboy or @kathys) knows if these fill values:

a. Still exist in the MiRS files
b. Are causing odd images in AWIPS

please let me know. If we can't find a reproducing case with modern versions of MiRS then I think we can close this. @kathys do you generate MiRS SFR geotiffs in production? Have you ever seen something like this?

spruceboy:
[...]

PR #304, opened 2021-02-16
Add MiRS reader

joleenf:
Part of change to address update of legacy MiRS reader mentioned in #269 and #151 might need checking.  Like the old reader, the new reader is only checking for missing values...

1. Adds a MiRS stub to the readers directory to work in combination with upcoming addition of MiRS reader in satpy.
2. Stub uses aliases similar to viirs_sdr.py
3. Default aliases checked in glue.py to existence of data id name in Scene object.  This adds a function which extracts the data name from the alias DataQuery and compares against the data names returned by the Scene.available_dataset_ids.  This should work even when readers do not use product aliases in the stub file, or if the list of product aliases contain strings.

glue checked with MiRS files for gpm and noaa satellites.  Glue checked with abi_l1b test files as well.

### DIFF
--- a/polar2grid/glue.py
+++ b/polar2grid/glue.py
@@ -32,7 +32,6 @@
 
 import os
 import sys
-import inspect
 import argparse
 import logging
 import importlib
@@ -41,7 +40,8 @@ from glob import glob
 
 import dask
 import numpy as np
-from pyresample.geometry import DynamicAreaDefinition, AreaDefinition
+from satpy import DataQuery
+from pyresample.geometry import DynamicAreaDefinition, AreaDefinition, SwathDefinition
 from pyproj import Proj
 from polar2grid.writers import geotiff, awips_tiled
 
@@ -87,7 +87,8 @@ def get_platform_name_alias(satpy_platform_name):
 def overwrite_platform_name_with_aliases(scn):
     """Change 'platform_name' for every DataArray to Polar2Grid expectations."""
     for data_arr in scn:
-        if 'platform_name' not in data_arr.__dict__:
+        data_dict = data_arr.attrs
+        if 'platform_name' not in data_dict:
             continue
         pname = get_platform_name_alias(data_arr.attrs['platform_name'])
         data_arr.attrs['platform_name'] = pname
@@ -172,8 +173,20 @@ def write_scene(scn, writers, writer_args, datasets, to_save=None):
         # no datasets to save
         return to_save
 
+    for dataID in datasets:
+        try:
+            area_def = getattr(scn[dataID], 'area', None)
+        except KeyError:
+            msg = 'No area definition in {}'.format(dataID)
+            LOG.info(msg)
+            LOG.error('Information', exc_info=True)
+            
+        if isinstance(area_def, SwathDefinition): 
+            scn[dataID].attrs['area'].area_id = "native"
+
     for writer_name in writers:
         wargs = writer_args[writer_name]
+        print(writer_name, datasets, wargs)
         res = scn.save_datasets(writer=writer_name, compute=False, datasets=datasets, **wargs)
         if isinstance(res, (tuple, list)):
             to_save.extend(zip(*res))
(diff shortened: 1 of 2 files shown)


polar2grid/glue.py
```diff
@@ -171,9 +174,20 @@ def write_scene(scn, writers, writer_args, datasets, to_save=None):
...
+            LOG.info(msg)
+            LOG.error('Information', exc_info=True)
+            
+        if isinstance(area_def, SwathDefinition): 
+            scn[dataID].attrs['area'].area_id = "native"
+
     for writer_name in writers:
         wargs = writer_args[writer_name]
-
+        print(writer_name, datasets, wargs)
```
djhoese, line 190: Extra print

polar2grid/glue.py
```diff
@@ -184,15 +198,18 @@ def write_scene(scn, writers, writer_args, datasets, to_save=None):
 
 def _handle_product_names(aliases, products):
     for prod_name in products:
-        yield aliases.get(prod_name, prod_name)
+        product = aliases.get(prod_name, prod_name)
+        if isinstance(product, DataQuery):
+            product = product['name']
```
djhoese, line 203: This is dangerous and we should be careful here. The DataQuery that this alias is using could include parameters other than name. For example, `band1_1000m` could be the 1000m resolution version of the MODIS band 1. That would be `DataQuery(name='1', resolution=1000)`. With this code that resolution information would get lost, right?

polar2grid/readers/mirs.py
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env python3
+# encoding: utf-8
+# Copyright (C) 2012-2015 Space Science and Engineering Center (SSEC),
```
djhoese, line 3: Update the years?

polar2grid/readers/mirs.py
```diff
@@ -0,0 +1,225 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# This file is part of the polar2grid software package. Polar2grid takes
+# satellite observation data, remaps it, and writes it to a file format for
+#     input into another program.
+# Documentation: http://www.ssec.wisc.edu/software/polar2grid/
+#
+# Written by Joleen Feltz    January 2021
```
djhoese, line 24: Now that there is more than one person contributing to this project :wink: maybe we should just drop this whole last portion. I find that they become out of date or don't represent the most recent authorship/maintainership of a module.

joleenf: Oh Yes.  I can drop that.  I was wondering about that.

polar2grid/readers/mirs.py
joleenf: Sure.

polar2grid/glue.py
joleenf: Yes it would.  This is interesting.  I need to think about that.  That could be handled if every reader handled the name concatenation in the exact same way, but that could be a bad assumption as well.

polar2grid/glue.py
joleenf: Found it!  I thought I had gotten rid of all of those...

polar2grid/glue.py
djhoese: You mean the way you are concatenating the polarizations onto the variable name? Yeah that is definitely not how other readers work. Most readers don't have this "compound" name like MiRS does. The DataQuery objects **need** to be passed to the Scene's load. Another easy example is the `calibration` parameter. I think some of my VIIRS code has aliases where "i04_radiance" can be loaded an is an alias for `DataQuery(name='I04', calibration='radiance')`.

joleenf (commented):
The changes are working.  I did a quick check of gpm product names and those I believe need to be added to the default names, but that can be handled later when I have a little more time to look a the files.

djhoese:
I'm impatient so I'm going to merge this so I can resolve some of the differences with my #305 PR.

PR #305, opened 2021-02-16
Add ability to configure resampling settings in YAML

djhoese:
This pull request adds a "decision tree" class for resampling similar to Satpy's enhancement logic. This means that a user (and Polar2Grid/Geo2Grid) can configure in a YAML file what resampler method and keyword arguments should be used to resampling a product. The choice will be made based on the product name, the sensor/instrument, the standard_name, and the type of area definition ("swath" or "area").

This pull request also includes major refactoring of the glue script to make these changes easier to digest.

This is overall needed so we can simplify the logic between Polar2Grid and Geo2Grid and provide a decent/flexible replacement for legacy Polar2Grid's argument parsing default resampling algorithms.


joleenf (approved):

djhoese:
There are some small hiccups with awips_tiled/scmi writing still, but this PR has grown out of scope so I'm going to merge this and start a new one to address the awips stuff.

Issue #268, opened 2020-03-25
Add SZA filtering to new glue interface


djhoese:
In order to get all Polar2Grid readers moved over to using the new Satpy-based glue script interface we need to port the various filters used in legacy P2G. The main one is the one that takes the solar zenith angles and says "don't generate this product because it is night time". I think this can be done strictly based on platform, sensor, and start_time using some of the utilities in satpy/pyresample/pyorbital.

One big thing to decide though:
Geo2Grid doesn't have these filters yet. Polar2Grid does and they are enabled by default. When they are ported to the satpy-based scripts, do we turn them on by default for both projects? P2G only? Is it confusing for users that these two similar projects will have different behavior when it comes to this?

kathys:
I think it is okay that G2G images are not SZA filtered.  When you think of a typical use case for making Geo imagery, it is as part of a loop.  I think you would prefer to have a blank image in your animation, than no image at all.  I will ask Graeme about this too to see what he would prefer.

djhoese:
I'm adding this as part of #308. The legacy P2G flags `--sza-threshold` is still there and that determines where the transition point (in solar zenith angle) is between day/night. Given that we don't want G2G to have this on by default, Kathy and I had talked about maybe making this off by default for P2G too. To support that the `--day-fraction` and `--night-fraction` flags (in legacy P2G) will now be called `--filter-day-products` and `--filter-night-products`. They can be specified with a fraction value:

```
--filter-day-products 0.2
```

Or with no value where it will default to 10% (0.1):

```
--filter-day-products
```

PR #308, opened 2021-02-22
Add day/night filtering using SZA

djhoese:
Porting some functionality from legacy Polar2Grid to Polar2Grid 3.0 with pyresample features instead of forcing SZA to be loaded.

This adds a `--filter-day-products` and `--filter-night-products` flag to control if filtering based on amount of day/night should be done and how much day/night is needed for a day-only/night-only product to be generated (as a fraction of the overall Scene). This also uses a `--sza-threshold` flag to control where the transition point is between day and night as the angle in degrees (default 100).

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,18 @@
+# Releasing the Polar2Grid Python Package
+
+1. Update setup.py with the new version.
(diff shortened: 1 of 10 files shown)


djhoese:
Closes #268

PR #314, opened 2021-03-10
Add --grid-coverage flag to avoid processing near-empty scenes

### LINKED ISSUE #312: Deprecate `--grid-coverage` flag in Polar2Grid
We need to discuss whether or not to include this flag in Polar2Grid 3.0. Polar2Grid <3.0 currently offers a `--grid-coverage X` flag which says "if the output grid doesn't have X% of the grid filled with data, don't output the file. This is a problem for P2G 3.0 because we are using dask which operates on array **chunks**. We can't know if the entire grid is covered unless we compute the entire grid and hold it all in memory at once. This removes all benefits of dask for the second half of process (mostly writing to disk) and would use a lot of memory.

One option would be to use polygon math based on only the lon/lats, similar to how day/night coverage is computed, to see what the coverage is before hand. I'm not entirely sure how to do this, but I'll see.

Just wanted to create this to make sure it was in the discussion for P2G 3.0.

djhoese:
Closes #312 by actually adding the flag back rather than deprecating it.

As discussed in #312, computing how many values of the output grid are actually filled with valid data points is not performant and defeats the purpose of using dask. The solution implemented here only loads the navigation data and does a check based on the bounding polygon of the input data and target grid to detect if the output will have data in it.

@kathys NOTE: This will not help in cases where, for example, the night portion of a reflectance channel is the only thing showing in the output grid. The output will still be produced. In past versions of P2G we were counting actual valid pixels. Here we are only checking coverage (swath/grid overlap). We *could* try to combine the logic used by the day/night filtering with the filtering implemented in this pull request, but not sure it is worth the effort right now. That still would only help with "invalid at night" products and not just bad data.


djhoese:
Closes #315 too.

Issue #315, opened 2021-03-11
Filter Key Failure


joleenf:
@djhoese I encountered an error when running the mirs reader with updates from master branch (including filters).  The error is due to a "day_only" KeyError:
```python
Traceback (most recent call last):
  File "/Users/joleenf/code/polar2grid/polar2grid/glue.py", line 550, in <module>
    sys.exit(main())
  File "/Users/joleenf/code/polar2grid/polar2grid/glue.py", line 519, in main
    scn = filter_scene(scn, reader_names,
  File "/Users/joleenf/code/polar2grid/polar2grid/filters/_filter_scene.py", line 91, in filter_scene
    criteria = get_reader_filter_criteria(reader_names, 'day_only')
  File "/Users/joleenf/code/polar2grid/polar2grid/filters/_filter_scene.py", line 52, in get_reader_filter_criteria
    readers_criteria = [_get_single_reader_filter_criteria(reader, filter_name)
  File "/Users/joleenf/code/polar2grid/polar2grid/filters/_filter_scene.py", line 52, in <listcomp>
    readers_criteria = [_get_single_reader_filter_criteria(reader, filter_name)
  File "/Users/joleenf/code/polar2grid/polar2grid/filters/_filter_scene.py", line 47, in _get_single_reader_filter_criteria
    return filter_criteria[filter_name]
KeyError: 'day_only'
```

Perhaps adding:

```python
FILTERS = { 'day_only': {}, 'night_only': {}, }
```

Is the only fix that is needed for the mirs reader?  However, there are probably other readers that will produce this error without the FILTERS line?

djhoese:
I ran into this too. I'm fixing it in #314

djhoese:
Closed by #314.

PR #316, opened 2021-03-11
Fix FILTER key failure in mirs reader as part of issue #315

joleenf:
Addresses KeyError noted in issue #315.


joleenf:
I am closing this because it is probably better to fix the filter_key error in one location rather than in every reader stub file.

djhoese:
As mentioned in #315, this is being fixed in #314.

PR #321, opened 2021-03-18
Add Satpy-compatible 'hdf5' writer

joleenf:
This is the start of the hdf writer.  The satpy code which executes main directly will disappear.  I am just using it to gain an understanding of this works.


polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,74 @@
...
+            if k in kwargs or k in self.info:
+                self.h5py_options[k] = kwargs.get(k, self.info[k])
+
+    @classmethod
+    def separate_init_kwargs(cls, kwargs):
+        """Separate the init keyword args."""
+        # FUTURE: Don't pass Scene.save_datasets kwargs to init and here
+        init_kwargs, kwargs = super(hdf5writer, cls).separate_init_kwargs(
+            kwargs)
+        for kw in ['dtype']:
```
djhoese, line 55: I thought of this in the middle of the night last night...if you don't need to have arguments in the `__init__` then I wouldn't define any and would not define this `separate_init_kwargs` method. The `__init__` versus `__call__` argument handling is one of the biggest warts in Satpy's writers. If we can avoid dealing with it later (by not defining `__init__` args) then let's do it.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def add_writer_argument_groups(parser, group=None):
+    """Create writer argument groups."""
+    from argparse import SUPPRESS
+    yaml = os.path.join(p2g_etc, "writers", "hdf5.yaml")
+    if group is None:
+        group = parser.add_argument_group(title='hdf5 Writer')
+    group.add_argument('--output-filename', dest='filename',
+                       default=DEFAULT_OUTPUT_PATTERN,
```
djhoese, line 261: I'd remove this default as glue will set it itself.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,292 @@
...
+
+from polar2grid.core.script_utils import (
+    setup_logging, rename_log_file, create_exc_handler)
+
+USE_POLAR2GRID_DEFAULTS = bool(int(os.environ.setdefault("USE_POLAR2GRID_DEFAULTS", "1")))
+
+LOG = logging.getLogger(__name__)
+
+# reader_name -> filename
+DEFAULT_OUTPUT_PATTERN = "{platform_name}_{sensor}_{start_time:%Y%m%d_%H%M%S}.h5"
```
djhoese, line 33: With current master I think this has to be `DEFAULT_OUTPUT_FILENAMES = {None: "{platform_name}..."}``

joleenf:
Not confident in save_datasets type hint, will it always get a List[xr.DataArray]?  Are there other possibilities like xr.DataArray, or List[da.Array] assuming xarray is imported as xr and dask as da.

djhoese:
> Not confident in save_datasets type hint, will it always get a List[xr.DataArray]? Are there other possibilities like xr.DataArray, or List[da.Array] assuming xarray is imported as xr and dask as da.

Always a list of DataArrays. Also, if you put `from __future__ import annotations` you can remove the `List` import and just use `list[xr.DataArray]`.

djhoese:
Did you run black on this?

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,302 @@
...
+
+dist = pkg_resources.get_distribution("polar2grid")
+if dist_is_editable(dist):
+    p2g_etc = os.path.join(dist.module_path, "etc")
+else:
+    p2g_etc = os.path.join(sys.prefix, "etc", "polar2grid")
+
+config_path = satpy.config.get("config_path")
+if p2g_etc not in config_path:
+    satpy.config.set(config_path=config_path + [p2g_etc])
```
djhoese, line 54: This should all be removed. If you need to use this logic in your tests then you should use `add_polar2grid_config_paths` from the glue script.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+def all_equal(iterable: list[str]) -> bool:
+    "Returns True if all the elements are equal to each other"
+    g = groupby(iterable)
+    return next(g, True) and not next(g, False)
+
+
+class FakeHDF5:
+    def __init__(self, fh : TextIO[IO[AnyStr]], output_filename: str, var_name: str, compression: bool):
```
djhoese, line 64: One of the important parts of this workaround (passing this FakeHDF5 instance) is that we don't pass anything that resembles an open file to dask. If I'm reading this correctly `fh` is the actual h5py File object, right? Is that needed? Can the dataset be created before hand?

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,302 @@
...
+        self.fh[self.var_name][write_slice] = data
+
+
+class hdf5writer(ImageWriter):
+    """Writer for hdf5 files."""
+
+    def __init__(self, dtype=None, **kwargs):
+        """Init the writer."""
+        super(hdf5writer, self).__init__(**kwargs)
+        self.dtype = self.info.get("dtype") if dtype is None else dtype
```
djhoese, line 87: Just to avoid unnecessary complexity in Satpy in the future, can you make `dtype` an argument on `save_datasets`?

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,302 @@
...
+            parent_group = self.create_proj_group(filename, hdf5_fh, area)
+
+            if add_geolocation:
+                chunks = data_arrs[0].chunks
+                dsets, target_fh = self.write_geolocation(
+                    hdf5_fh, filename, parent_group, area, append, compression, chunks
+                )
+            else:
+                dsets = list()
+                target_fh = list()
```
djhoese, line 254: I could have my indentation wrong, but aren't you overwriting the `dsets` and `target_fh` for every for loop? So only one area would ever be possible?

djhoese (changes requested):
Nice job. It is really close I think. I have a few concerns that I've made comments for inline.

polar2grid/writers/hdf5.py
joleenf: Okay.  I am a little confused on this one.  Yes, I can pass the dask array and the closed file handle to dask?  Or, are you saying I should pass the dask array and the hdf5 variable name to dask?  Though we are trying to avoid passing an open file handle, we are trying to pass a dask array that is either delayed or computed, correct?

djhoese: The `FakeHDF5` file `__init__` method should only get non-dask and non-open-file objects so strings (filenames, variable names, etc) and other configuration options like booleans. We want to avoid `FakeHDF5` holding on to open file objects (like the h5py File object or other h5py objects) as they can't be serialized which dask will need to do when passing it around in certain situations. Computed numpy arrays would be fine, but I'm not sure you need them in this writer.

The `__setitem__` is a different story as dask will already have serialized and computed things for you.

polar2grid/writers/hdf5.py
joleenf: Should be fixed.  b56af2cc

polar2grid/writers/hdf5.py
joleenf: I think I have this resolved in b56af2cc.  The FakeHDF5 class constructor only accepts a filename and var_name.  The ```__setitem__``` writes by opening a file handle to the filename which already has the datasets defined, just not filled with data.  b56af2cc

polar2grid/writers/hdf5.py
joleenf: Urgh!  That definitely could happen.

joleenf: b021a9 defines a new list for each AREA group.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+    def __init__(self, output_filename: str, var_name: str):
+        self.output_filename = output_filename
+        self.var_name = var_name
+
+    def __setitem__(self, write_slice, data):
+        with h5py.File(self.output_filename, mode="a") as fh:
+            fh[self.var_name][write_slice] = data
+
+
+class hdf5writer(ImageWriter):
```
djhoese, line 51: Would you mind updating this to `HDF5Writer` for my own style-ish needs. YAML will need to be updated too.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+USE_POLAR2GRID_DEFAULTS = bool(int(os.environ.setdefault("USE_POLAR2GRID_DEFAULTS", "1")))
+
+LOG = logging.getLogger(__name__)
+
+# reader_name -> filename
+DEFAULT_OUTPUT_FILENAMES = {None: "{platform_name}_{sensor}_{start_time:%Y%m%d_%H%M%S}.h5"}
+
+
+def all_equal(iterable: list[str]) -> bool:
+    "Returns True if all the elements are equal to each other"
```
djhoese, line 36: Three quotes and a period at the end. I'm surprised black doesn't complain about/fix this.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+        args = dataset.attrs
+        args["grid_name"] = "native" if isinstance(area, SwathDefinition) else area.area_id
+        args["rows"], args["columns"] = area.shape
+        args["data_type"] = dtype
+
+        return args
+
+    def iter_by_area(self, datasets: list[xr.DataArray]):
+        """Generate datasets grouped by Area.
+        :return: list of (area_obj, list of dataset objects)
```
djhoese, line 75: I don't think this is documented anywhere, but I'd like Polar2Grid to follow "google style" docstrings like Satpy. So return should be:

```
"""Generate datasets grouped by Area.

Returns:
    list of (area_obj, list of dataset objects)
```

djhoese: You could also try to add type annotations for the return value instead (might be ugly).

```
) -> list[tuple[AreaDefinition, list[DataArray]]]:
```

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+        """Create a filename for saving output data.
+
+        Args:
+            kwargs (dict): Attributes and other metadata to use for formatting
+                the previously provided `filename`.
+
+        """
+        if self.filename_parser is None:
+            raise RuntimeError("No filename pattern or specific filename provided")
+        self.filename_parser = Parser(convert_p2g_pattern_to_satpy(self.filename_parser.fmt))
```
djhoese, line 93: Could this conversion happen in the `__init__` of the writer? That way you don't have to subclass this whole method, right?

Or, if we assume that all patterns configured for the writer are current/correct, then we could do the conversion on the `--output-filename` command line argument. Thoughts?

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+            LOG.info("Creating HDF5 file: %s", output_filename)
+            mode = "w"
+
+        h5_fh = h5py.File(output_filename, mode)
+        return h5_fh
+
+    @staticmethod
+    def create_proj_group(filename: str, parent: TextIO, area_def):
+        """Create the top level group from projection information."""
+        projection_name = (area_def.name).replace(" ", "_")
```
djhoese, line 121: I *think* this needs to be `area_def.area_id`.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+        lon_dset = self.check_variable(fh, lon_grp, lon_data, dtype, append, compression)
+        lat_dataset = FakeHDF5(fname, lat_grp)
+        lat_dset = self.check_variable(fh, lat_grp, lat_data, dtype, append, compression)
+
+        return [lon_data, lat_data], [lon_dataset, lat_dataset]
+
+    @staticmethod
+    def create_variable(filename: str, parent, hdf_subgroup: str, dataset_id: xr.DataArray, dtype: np.dtype, **kwargs):
+        """Create a hdf5 subgroup and it's attributes."""
+
```
djhoese, line 176: Can you remove this blank line? Kind of surprised black doesn't catch this either.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+        dataset: list[xr.DataArray],
+        filename=None,
+        dtype=None,
+        fill_value=None,
+        append=True,
+        compute=True,
+        **kwargs,
+    ):
+        """Save hdf5 datasets."""
+
```
djhoese, line 208: Same thing with the blank line here.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+        output_names = []
+
+        for dataset_id in dataset:
+            args = self._output_file_kwargs(dataset_id, dtype)
+            out_filename = filename or self.get_filename(**args)
+            output_names.append(out_filename)
+        one_file = all_equal(output_names)
+
+        filename = output_names[0]
+        if not all_equal(output_names):
```
djhoese, line 225: Seems like some duplicate usage of `all_equal` here. I'm guessing the `one_file =` line doesn't need to exist?

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+            LOG.warning("Writing to {}".format(filename))
+
+        hdf5_fh = self.open_hdf5_filehandle(filename, append=append)
+
+        datasets_by_area = self.iter_by_area(dataset)
+        for area, data_arrs in datasets_by_area:
+            # Initialize source/targets at start of each new AREA grouping.
+            dsets = []
+            targets = []
```
djhoese, line 234: I think this is still wrong. If there is more than one area group then `targets` and `dsets` is overwritten when the second group is being processed and `return (dsets, targets)` at the bottom of this method will never include the earlier group(s).

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+
+            for data_arr in data_arrs:
+                d_dtype = data_arr.dtype if dtype is None else dtype
+                hdf_subgroup = "{}/{}".format(parent_group, data_arr.attrs["name"])
+
+                file_var = FakeHDF5(filename, hdf_subgroup)
+                arr_dset = self.check_variable(hdf5_fh, hdf_subgroup, data_arr.data, d_dtype, append, compression)
+                dsets.append(data_arr.data)
+                targets.append(file_var)
+        return (dsets, targets)
```
djhoese, line 254: Redundant parentheses.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+        choices=NumpyDtypeList(NUMPY_DTYPE_STRS),
+        type=str_to_dtype,
+        help="Data type of the output file (8-bit unsigned " "integer by default - uint8)",
+    )
+    group.add_argument(
+        "--fill-value", dest="fill_value", type=int_or_float, help="Fill value for invalid data in this dataset."
+    )
+    group.add_argument("--compress", default="LZW", help="File compression algorithm (DEFLATE, LZW, NONE, etc)")
+    group.add_argument(
+        "--chunks",
```
djhoese, line 275: I don't think you ever use this, right?

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,293 @@
...
+        dest="chunks",
+        type=tuple,
+        help="Chunked storage of dataset. " "Default is chosen appropriate to dataset by h5py.",
+    )
+    group.add_argument(
+        "--add-geolocation",
+        dest="add_geolocation",
```
djhoese, line 282: Probably my fault, but this `dest=` is unnecessary in this argument.

djhoese (changes requested):
A couple questions and a couple requests. Looking really good though.

polar2grid/writers/hdf5.py
joleenf: Are you referring converting in glue at _parse_writer_args? (Assuming all patterns configured are correct).

polar2grid/writers/hdf5.py
joleenf: Correct

polar2grid/writers/hdf5.py
joleenf: Correct, however, will this be a command line argument for the future?  Should I actually try to use a user input chunk size, or is P2G always going to use a default chunk size?

polar2grid/writers/hdf5.py
joleenf: Direct copy, I am sure there is split opinion regarding blame...at least 50% my fault.  Not a problem, this is fixable.

polar2grid/writers/hdf5.py
djhoese: I was more thinking of specifying the `type` in the `add_argument` to be a function that does the conversion. You can see this idea in the `int_or_float` usage in the command line arguments for your writer already (or at least `--fill-value` was using it).

polar2grid/writers/hdf5.py
djhoese: I think this is referring to the HDF5 chunk size (internal to the HDF5 file). Especially if this came from the old writer then this has nothing to do with dask (we didn't use dask then).

joleenf: My question is directed more towards using dask chunk sizes.  Correct me if I misunderstand, if I send the chunk size to da.store, won't dask control the chunk size written to the HDF5 file?  If that is the case, did you want to see an option in the future for controlling that chunk size, or just use the default chunk size?

djhoese: There will likely not be a command line argument to control dask array chunk size. If anything, it will be an environment variable (inline with Satpy's 'Config' object and its usage). If you'd like to rename this to `--hdf5-chunks` to avoid confusion I am ok with that.

Oh! I just reread the old writer code. You added this `--chunks` didn't you? In that case, please remove this as an option. I thought this was controlling the chunk size of the HDF5 variables on disk.

joleenf:
> What happens if this isn't done? Since this is using the new `start_time` name (as opposed to `begin_time`) then I would think this is what the user wants (whatever the output is).

Yes, that could be true..it is ugly output because the attributes have the isoformat time for use in the HDF5  file attributes.   But you are correct, for some writers, the default may be perfect.

djhoese:
> for some writers, the default may be perfect.

**NO** new writers should use legacy parameters as possible default output filenames. Only new-style parameters.

joleenf:
I didn't write what I meant.  The new output-pattern should have the new satpy kwargs, but if the default format for start-time is start_time.isoformat() in HDF5, it may be start_time.strftime("%Y%m%d_%H%M%S") in another and that later would be just right imo.  The HDF5 default time format it kind of ugly for a file name.  In the spirit of keeping the attributes in the HDF file the same, that was transferred directly from the original writer.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,283 @@
...
+    def __init__(self, output_filename: str, var_name: str):
+        self.output_filename = output_filename
+        self.var_name = var_name
+
+    def __setitem__(self, write_slice, data):
+        with h5py.File(self.output_filename, mode="a") as fh:
+            fh[self.var_name][write_slice] = data
+
+
+class HDF5writer(Writer):
```
djhoese, line 45: Sorry, capital 'W'

polar2grid/utils/legacy_compat.py
```diff
@@ -25,6 +25,7 @@
 from __future__ import annotations
 
 import logging
+import re
```
djhoese, line 28: Not used anymore right?

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def add_writer_argument_groups(parser, group=None):
+    """Create writer argument groups."""
+    from argparse import SUPPRESS
+
+    if group is None:
+        group = parser.add_argument_group(title="hdf5 Writer")
+    group.add_argument(
+        "--output-pattern",
```
djhoese, line 259: I think to match the new writers this should be `--output-filename`.

joleenf: There will be no backwards compatibility with this?  If someone enters --output-pattern, it should fail?

joleenf: Or should I be nice and convert it.

djhoese: I was thinking no there wouldn't be backwards compatibility, but I'm not sure what @kathys and I decided on this in one of our recent meetings. If you try to convert to the new kwarg it would have to be handled by the glue script and not the writer as we'd have to do it for Satpy-centric writers too.

joleenf: @kathys @djhoese Not sure of your opinion, the argument parser will throw an error which is pretty clear about --output-pattern being the wrong argument.  What it can't do at the moment is redirect the user to using the --output-filename flag.  However, this could get ugly over time if multiple old flags need to throw individualized messages.  Either, leaving with the simple error message or printing the default help from the parser might be the better than trying anything more customized.  @djhose, your opinion?  Am I walking down a path that is too hard to support?

djhoese: Similar to have the `type=` defined for the argument as a function that does the legacy keys -> new keys conversion, we could have the function print a message telling them the new flag to use *or* raise an error telling them the new flag to use. The help message could also be updated to mention the new flag.

Is any of this necessary...I don't know. Is it easier to just make users learn the new way? Maybe?

polar2grid/utils/legacy_compat.py
```diff
@@ -56,6 +57,10 @@ def convert_old_p2g_date_frmts(frmt):
     return frmt
 
 
+def output_pattern_parser_error(value):
+    raise DeprecationWarning("--output-pattern is deprecated, use --output-filename")
```
djhoese, line 61: Interesting. I've never seen warnings used this way before. Usually it is `warnings.warn("some message", DeprecationWarning)`, but maybe in this case this is the right approach? Do the python docs talk about this?

joleenf: Looking at the docs, it probably shouldn't be used that way.  However, it seems silly to keep moving forward.  However, the code will raise an argument parse exception later, so it is proabably fine just to use warnings.warn as was intended with DeprecationWarning (not DeprecationError).

djhoese: Note warnings.warn does not print out DeprecationWarnings by default. Use the default UserWarning if you want that.

polar2grid/writers/hdf5.py
joleenf: I am going to back track on this, if error doesn't get raised early, argparse doesn't actually throw the error for output-patter, because it actually does exist.  So it is better to just let argparse throw and error and then let the user go back to the docs. 
```
usage: 
    /Users/joleenf/code/polar2grid/polar2grid/glue.py -h
see available products:
    /Users/joleenf/code/polar2grid/polar2grid/glue.py -r <reader> -w <writer> --list-products -f file1 [file2 ...]
basic processing:
    /Users/joleenf/code/polar2grid/polar2grid/glue.py -r <reader> -w <writer> [options] -f file1 [file2 ...]
basic processing with limited products:
    /Users/joleenf/code/polar2grid/polar2grid/glue.py -r <reader> -w <writer> [options] -p prod1 prod2 -f file1 [file2 ...]
/Users/joleenf/code/polar2grid/polar2grid/glue.py: error: unrecognized arguments: --output-pattern

```

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,327 @@
...
+        # will this be written to one or multiple files?
+        output_names = []
+
+        for dataset_id in dataset:
+            file_attrs = self._output_file_kwargs(dataset_id, dtype)
+            out_filename = filename or self.get_filename(**file_attrs)
+            output_names.append(out_filename)
+
+        filename = output_names[0]
+        if not all_equal(output_names):
```
djhoese, line 253:
```suggestion
        if not all(output_names[0] == output_name in output_name in output_names[1:]):
```

Potential "shorter" version that doesn't require additional imports.

djhoese: Or I guess since you did `filename = ` you could do `filename == output_name` in the loop.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,327 @@
...
+            # This writer had only delayed writes
+            return delayeds
+        else:
+            return targets, sources
+
+
+def add_writer_argument_groups(parser, group=None):
+    """Create writer argument groups."""
+    if group is None:
+        group = parser.add_argument_group(title="hdf5 Writer")
```
djhoese, line 300:
```suggestion
        group = parser.add_argument_group(title="HDF5 Writer")
```

polar2grid/writers/hdf5.py
djhoese: I changed the implementation of `all_equal` to use this technique. I had a hard time following how it was working in `all_equal`. I also changed the warning message to be more clear about why it is a warning.

polar2grid/writers/hdf5.py
```diff
@@ -0,0 +1,326 @@
...
+#
+# Written by David Hoese    December 2014
+# University of Wisconsin-Madison
+# Space Science and Engineering Center
+# 1225 West Dayton Street
+# Madison, WI  53706
+# david.hoese@ssec.wisc.edu
+
+"""
+The HDF5 writer creates HDF5 files with data arrays from satpy.Scene.
```
djhoese, line 32:
```suggestion
"""The HDF5 writer creates HDF5 files with groups for each gridded area.
```

djhoese (approved):
I changed a few nitpicks of my own, but otherwise looks good. I'll merge it once tests pass.

Issue #326, opened 2021-03-25
Polar2Grid Version 3.0 issues


kathys:
Here is a checklist of the issues that I found with the Version 3.0 tarballs.

### Tarball: `polar2grid-swbundle-20210324-121314.tar.gz`

- [x] The false color and true color image is being created using an **ewa D=40 d=1**.  It should be **D=40 and d=2**. See #325 for further discussion.
- [x] The viirs_sdr reader lists the band options (_--list-products_) in upper case.  It was previously lower case.
- [x] Background of GeoTIFFs is white (probably transparent).  It used to be black.  We should discuss this and its potential effects on users.
- [x] Doing `source $POLAR2GRID_HOME/bin/polar2grid_env.sh` fails.

**Edit by Dave**: I've added a section header with the name of the tarball. As we come out with new tarballs that may have new and different issues we can add new sections to keep track of what tarball causes or fixes what issues.

### Tarball: `polar2grid-swbundle-20210422-181315.tar.gz`

- [x] creating multiple grids from a grid-configs file using a single polar2grid command creates only one of the grids but duplicates it for all of the requested grids.  The output files have the right names, but they are just duplicates of the last grid in the list.  Example: $POLAR2GRID_HOME/bin/polar2grid.sh -r viirs_sdr -w geotiff -p true_color false_color --num-workers 4 --tiled --progress --grid-configs ../test_grids.conf -g lake_winnipeg hot_springs grand_rapids spirit_lake white_sands campo_dos_campo_mx texas -f /data/us_sdr
- [x] Threshold for making visible images in predominately nighttime scenes is different than the value used in old p2g
- [x] White background visible band images all show some values in non-solar illuminated section of data.  See montage below.  The band order is from top to bottom, right to left, i01-i05, m01-m16.  The infrared band images are easy to spot - they are the ones that are consistent  throughout the pass.

![noaa20_viirs_20210601_081815_montage](https://user-images.githubusercontent.com/1848019/120392219-cd2d3480-c2f5-11eb-8e6e-410502e601ae.png)

- [x]  First and last lines of images are not complete.  Below is an example close up of a VIIRS DNB image upper right corner.  Right panel is image from old P2G, and left is from new P2G.
        Test data for this example is in:  bumi:/data/test_data/old_polar2grid_data/viirs_sdr_2021/night/input
         P2G V3.0 Images and log file in:  bumi:/data/test_data/old_polar2grid_data/viirs_sdr_2021/night/new_fill0
[...]

Issue #325, opened 2021-03-25
True Color image differences


kathys:
Installed the polar2grid-swbundle-20210324-121314.tar.gz tarball and created a true color image from a test data set that I have on _bumi:/data/users/kathys/test_data/viirs_asia/sdr_.  When I compare with the original images that I made using Polar2Grid Version 2.3 I noticed these differences

- [ ]  The background is white (or transparent). It was previously black.
- [ ]  The images is sharper.
- [ ]  There are "holes" in the image.
- [ ]  The enhancement is different.

Here are full resolution cutouts from a section of the image created using the P2G Version 2.3 and P2G Version 3.0 software.

Version 2.3
![P2G_v2 3](https://user-images.githubusercontent.com/1848019/112509404-6ec86080-8d5e-11eb-9846-d45ee1fafaba.png)
Version 3.0
![P2G_v3 0](https://user-images.githubusercontent.com/1848019/112509443-77b93200-8d5e-11eb-9d56-d61e2fc81a31.png)

djhoese:
@kathys Can you verify that the background isn't actually transparent? That's what I'm seeing in the images I'm generating. If you open the images in Preview on OSX there should be a Preferences option to change the background color and/or remove it. If you specify the `--fill-value 0` flag on the command line then you should get a black background. The default behavior (we've discussed this a couple months ago) is to create an "alpha" channel on these RGBs to control the transparency (transparent for invalid pixels).

kathys:
Verified that background is transparent.  I do remember talking about this; I just need to remember to put this info into the documentation.

yukaribbba:
I would say I prefer the 2.3 version. It will give me more latitude when processed by Photoshop or other image tools.

djhoese:
@yukaribbba Yes, we agree. This issue is meant to address issues/differences in the current internal beta version of P2G 3.0.

djhoese:
Just to tie up this issue, @kathys has done a lot of investigating into this and the images are now the same with current development versions of the software (polar2grid, satpy, trollimage, etc).

PR #329, opened 2021-03-26
Update VIIRS crefl usage to download CMGDEM.hdf file

djhoese:
This depends on newly added Satpy functionality. The Modifier in Satpy now downloads the CMGDEM.hdf file instead of defaulting to all 0s for the elevation dataset (if it can't find the file). This PR uses that functionality as well as the improved handling of putting a hard requirement on the angle datasets (previously optional which could cause issues in specific situations).


djhoese:
See related #325 and #326

PR #336, opened 2021-04-15
Rows update resample

joleenf:
As commit states, adds rows_per_scan for mirs sensors only.


djhoese:
Looks like you included your HDF5 writer in this PR. Any chance you could separate it out? Will probably require a force push or a new PR. Be careful not to lose your writer stuff.

joleenf:
I will just start a new branch from master and a new pull request.  I branched from the hdf_writer branch, that was the problem.

PR #337, opened 2021-04-15
Fix rows_per_scan not being defined for MiRS resampling

joleenf:
This adds a rows_per_scan=0 for mirs sensors to work with pyresample v 1.19 or higher.


etc/resampling.yaml
```diff
@@ -12,7 +12,7 @@ resampling:
     resampler: ewa
     kwargs:
       weight_delta_max: 40.0
-      weight_distance_max: 1.0
+      weight_distance_max: 2.0
```
djhoese, line 15: This is very confusing. This is already the value in master isn't it? https://github.com/ssec/polar2grid/blob/f21df1604e317d26ccf544905829c93fccc10b53/etc/resampling.yaml#L15

djhoese: Oh I think your master branch is behind the ssec/polar2grid master. Maybe that's why.

joleenf:
Yes.  It is behind.  That is why that difference appears.

joleenf:
Is it easier if I just update my master branch with upstream and push again?

djhoese (approved):
LGTM. Thanks.

PR #342, opened 2021-04-18
Change binary writer to default 64-bit floats to 32-bit floats

djhoese:
To avoid inconsistencies as Satpy changes, the binary writer needs to have consistent output for floating point data. With current versions of xarray (this may change in the future) it is very easy for Satpy to accidentally upcast 32-bit floats to 64-bit floats. This amount of precision is unnecessary for most remote-sensing satellite instruments and wastes space and processing time. This PR makes it so instead of defaulting to the same data type as the input data, the writer will cast 64-bit floats as 32-bit floats. This way if Satpy or xarray changes in the future produce the same data as 32-bit floats, this won't change output for P2G users.

This PR also includes fixing output filenames for the geotiff and binary writers, fixing the default radiance enhancement configuration, and adding a `--plot` option to the compare script for debugging.

### DIFF
--- a/doc/source/readers/viirs_sdr.rst
+++ b/doc/source/readers/viirs_sdr.rst
@@ -1,35 +1,29 @@
 VIIRS SDR Reader
 ================
 
-.. automodule:: polar2grid.viirs.swath
+.. automodule:: polar2grid.readers.viirs_sdr
 
 Command Line Arguments
 ----------------------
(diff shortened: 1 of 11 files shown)


djhoese:
This will keep failing until https://github.com/pytroll/satpy/pull/1646 is merged.

PR #346, opened 2021-04-24
Overhaul comparison scripts to provide summary HTML pages

djhoese:
In an attempt to make @kathys's life easier, this PR starts the work to include HTML summary pages for all of the integration tests running on our Jenkins server. The idea is that when this is fully implemented we can go to the jenkins page, click on the link to view the generated HTML page and see each tests success or failure, percentage of pixels that differed, and see thumbnails of the actual images that were generated and the ones that were expected.


djhoese:
In its current state this PR generates a very basic HTML file. With basic local tests I have shown that the HTML is generated fine. Now I just need to test the integration tests and make sure they present the HTML files properly. Rather than make multiple projects on the Jenkins server, I'm going to just merge this and see what happens.

PR #350, opened 2021-05-14
Fix various HDF5 writer issues

joleenf:
HDF5 writer: Metadata Fixes: change grid projection string to the proj4_string, reverses cell height and adds x and y origin values.  Write the p2g_name to the files rather than the satpy name for the variables.  

Update the pattern conversion so that begin_time defaults to the format that was produced in the previous output_filename which is now start_time:%Y%m%d_%H%M.  start_time, when entered alone should still output the default date pattern which is in isoformat.

### DIFF
--- a/polar2grid/utils/legacy_compat.py
+++ b/polar2grid/utils/legacy_compat.py
@@ -25,6 +25,7 @@
 from __future__ import annotations
 
 import logging
+import re
 
 from typing import Union, Iterable, Generator, Optional
 from satpy import Scene, DataID, DataQuery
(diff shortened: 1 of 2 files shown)


polar2grid/utils/legacy_compat.py
```diff
@@ -65,6 +63,13 @@ def convert_p2g_pattern_to_satpy(pattern):
         "begin_time": "start_time",
         "product_name": "p2g_name",
     }
+    # If there is no other formatting before replacing begin_time,
+    # add the old default formatting to the pattern and changed begin_time to
+    # start_time
+    abs_search = re.compile(r"{begin_time}").search
+    if bool(abs_search(pattern)):
```
djhoese, line 70: Are regular expressions needed? Or even this if statement? Would only doing:

```
pattern = pattern.replace("{begin_time}", "{start_time:%Y%m%d_%H%M}")
```

work?

djhoese (approved):
LGTM. Thanks for the fixes.

Issue #318, opened 2021-03-12
Re-add legacy P2G readers from Satpy implementations


djhoese:
The below list is every reader that we provided in the previous version of Polar2Grid and still want to support. While #269 deals with converting P2G legacy readers to a Satpy implementation, this issue deals with making sure P2G has access to Satpy readers to access reading functionality that was available previously. For example, Satpy already has a VIIRS SDR and VIIRS L1b reader, but these weren't available in the normal Polar2Grid interfaces and providing the same (or very similar) set of defaults and products. For completeness all readers are listed below:

- [x] VIIRS SDR
- [x] VIIRS L1b
- [x] MODIS L1b
- [x] AVHRR (AAPP)
- [x] AMSR2 L1b
- [x] NUCAPS
- [x] MiRS
- [x] ACSPO SST
- [x] CLAVR-x
- [x] VIIRS EDR Active Fires
- [x] VIRR L1b
- [x] MERSI-2 L1b

djhoese:
This accidentally got closed from #359. This is not finished.

djhoese:
This is finished now. We have further more in-depth testing to do, but the readers do exist in the master branch now.

PR #354, opened 2021-05-26
Add proxy module for Satpy clavrx reader

joleenf:
Update the Clavr-x wrapper referenced in #318.  A dictionary has been added with day/night products for filter following the viirs_sdr wrapper.  This means that the day/night product filter uses standard names rather than variable names.  Changes to wrapper are based on the example set in the viirs_sdr reader.  The Frontend class and group arguments have been removed.  The DAY_ONLY and NIGHT_ONLY variable names are still in the wrapper, but as far as I can tell, they are not used?  I am concerned that the standard_name for the  toa_bidirectional_reflectance is too general because channel 7 is a mixed solar and thermal and produces a product in both day and night hours.

### DIFF
--- a/polar2grid/readers/clavrx.py
+++ b/polar2grid/readers/clavrx.py
@@ -100,11 +100,6 @@ else:
         ]
     )
 
-DAY_ONLY = ["cld_opd_dcomp", "cld_reff_dcomp"]
-NIGHT_ONLY = ["cld_opd_nlcomp", "cld_reff_nlcomp", "refl_lunar_dnb_nom"]
-
-FILE_EXTENSIONS = [".hdf"]
-DEFAULT_READER_NAME = "clavrx"
 DEFAULT_DATASETS = [
     "cloud_type" "cld_temp_acha",
     "cld_height_acha",
@@ -146,16 +141,6 @@ class ReaderProxy(ReaderProxyBase):
     def get_default_products(self) -> list[str]:
         """Get products to load if users hasn't specified any others."""
         return set(DEFAULT_DATASETS) & self.get_all_products()
-
-    def filter(self, scn):
-        self.filter_daynight_datasets(scn)
-
-    def filter_daynight_datasets(self, scene):
-        """Some products are only available at daytime or nighttime"""
-        for k in DAY_ONLY + NIGHT_ONLY:
-            if k in scene and scene[k].isnull().all():
-                LOG.info("Removing dataset '{}' because it is completely empty".format(k))
-                del scene[k]
 
 
 def add_reader_argument_groups(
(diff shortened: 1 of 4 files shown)


polar2grid/readers/clavrx.py
```diff
@@ -64,73 +64,91 @@
...
+            "refl_lunar_dnb_nom",
+            "rain_rate",
+        ]
+    )
+
+DAY_ONLY = ["cld_opd_dcomp", "cld_reff_dcomp"]
+NIGHT_ONLY = ["cld_opd_nlcomp", "cld_reff_nlcomp", "refl_lunar_dnb_nom"]
+
+FILE_EXTENSIONS = [".hdf"]
+DEFAULT_READER_NAME = "clavrx"
```
djhoese, line 107: I think these can be removed.

polar2grid/readers/clavrx.py
```diff
@@ -64,102 +64,87 @@
...
+            "cld_reff_dcomp",
+            "cld_reff_nlcomp",
+            "cld_emiss_acha",
+            "refl_lunar_dnb_nom",
+            "rain_rate",
+        ]
+    )
+
+DEFAULT_DATASETS = [
+    "cloud_type" "cld_temp_acha",
```
djhoese, line 104: Looks like this was a typo in the original list. This should have a comma between it it seems.

polar2grid/readers/clavrx.py
```diff
@@ -64,102 +64,87 @@
...
+}
+
+
+class ReaderProxy(ReaderProxyBase):
+    """Provide Polar2Grid-specific information about this reader's products."""
+
+    is_geo2grid_reader = True
+    is_polar2grid_reader = True
+
+    def available_product_names(self) -> sorted(set(str)):
```
djhoese, line 134: I don't think this is a method for the ReaderProxyBase class. Also are you allowed to do a return type definition of `sorted(set(str))`?

polar2grid/readers/clavrx.py
```diff
@@ -64,102 +64,88 @@
...
+}
+
+
+class ReaderProxy(ReaderProxyBase):
+    """Provide Polar2Grid-specific information about this reader's products."""
+
+    is_geo2grid_reader = True
+    is_polar2grid_reader = True
+
+    def available_product_names(self) -> set(str):
```
djhoese, line 135: This can be removed, right?

djhoese: The whole method I mean. Also, I think the type hint is supposed to be `set[str]` but it doesn't matter if the method is removed.

joleenf:
@djhoese A crude stretch has been added to the cloud phase and cloud type products.  Please take note of that commit.  Without the stretch, the cloud phase was being stretch from 0-255 for the geotiff writer.

djhoese (approved):
Looks good to me. I'll merge this and any future updates can be made in another PR. Thanks @joleenf!

PR #355, opened 2021-06-10
Viirs l1b

joleenf:
Addresses #318 for VIIRS_L1b reader

Old frontend is removed, aliases updated  to use DataQuery, make sure p2g names are converted to lower case, adds new argument parser that glue will use.  Day/Night filters copied from VIIRS_SDR.


joleenf:
This is one of the questions I have:  I don't specifically see "natural color" as a product, only as a description under false color in the documentation.  
https://github.com/joleenf/polar2grid/blob/a1c15da8bc3f5cc8a6fa353826544000cd6834af/polar2grid/readers/viirs_l1b.py#L273

djhoese:
I think that's a typo. There was a time when I thought the "false color" that we had always been producing in P2G was the same as the "natural color" that Satpy was creating. They are not the same. This description problem came during this time.

djhoese:
Also, I had learned that false color is only something specific that a few MODIS folks in the building (like @kathys ) use, but in other organizations false color is just the generic term for not-true color and includes technically any RGB.

joleenf:
@djhoese This can be reviewed.  Do the filenames need to match old p2g output patterns?  There is a difference in how the reader presents the platform_id name to the writers, so the filenames do not match exactly.  fy3b vs. FY-3B now.  Should the conversion be handled in the wrapper?

etc/enhancements/viirs.yaml
```diff
@@ -1,4 +1,18 @@
...
+      - name: linear_stretch
+        method: !!python/name:satpy.enhancements.stretch
+        kwargs: { stretch: 'crude', min_stretch: 0.0, max_stretch: 1.0 }
+  histogram_dnb:
+    sensor: viirs
+    standard_name: equalized_radiance
+    operations:
+      - name: linear_stretch
+        method: !!python/name:satpy.enhancements.stretch
+        kwargs: { stretch: 'crude', min_stretch: 0.0, max_stretch: 1.0 }
```
djhoese, line 15:
```suggestion
  normalized_dnb:
    sensor: viirs
    standard_name: equalized_radiance
    operations:
      - name: linear_stretch
        method: !!python/name:satpy.enhancements.stretch
        kwargs: { stretch: 'crude', min_stretch: 0.0, max_stretch: 1.0 }
```

The section header doesn't actually mean anything beyond being a unique identifier/key. Having these two sections be the exact same thing ends up being redundant.

polar2grid/readers/viirs_l1b.py
```diff
@@ -27,24 +27,26 @@
 #     1225 West Dayton Street
 #     Madison, WI  53706
 #     david.hoese@ssec.wisc.edu
-"""The VIIRS Level 1B Reader operates on NASA Level 1B (L1B) NetCDF files 
-from the Suomi National Polar-orbiting Partnership's (NPP) Visible/Infrared
+"""
+The VIIRS Level 1B Reader operates on NASA Level 1B (L1B) NetCDF files.
```
djhoese, line 31:
```suggestion
"""The VIIRS Level 1B Reader operates on NASA Level 1B (L1B) NetCDF files.
```

polar2grid/readers/viirs_l1b.py
```diff
@@ -184,150 +190,150 @@
...
+        const=_AWIPS_TRUE_COLOR,
+        help="Add the True Color product to the list of products",
+    )
+    group.add_argument(
+        "--false-color",
+        dest="products",
+        action=ExtendConstAction,
+        const=_AWIPS_FALSE_COLOR,
+        help="Add the False Color product to the list of products",
+    )
```
djhoese, line 338:
```suggestion
    group.add_argument(
        "--awips-true-color",
        dest="products",
        action=ExtendConstAction,
        const=_AWIPS_TRUE_COLOR,
        help="Add the True Color product to the list of products",
    )
    group.add_argument(
        "--awips-false-color",
        dest="products",
        action=ExtendConstAction,
        const=_AWIPS_FALSE_COLOR,
        help="Add the False Color product to the list of products",
    )
```

djhoese:
While reviewing this PR I noticed how inconsistent the product naming was between the SDR reader and this one. I also noticed some issues with what the reader actually calls the products and what we were calling it here (or backwards incompatibility with what was done before). I think I've made everything as consistent as I can while having aliases in place for backwards compatibility. Thanks @joleenf for getting this set up.

PR #358, opened 2021-06-22
VIRR (fy3b, fy3c) Level 1 B wrapper

joleenf:
This can be reviewed.  I am wondering about default output filenames.  The old reader provided fy3b, fy3c to the output file pattern, the new reader if providing FY-3B, FY-3C, should this be handled in the wrapper, legacy compat or leave as is?


djhoese:
> The old reader provided fy3b, fy3c to the output file pattern, the new reader if providing FY-3B, FY-3C, should this be handled in the wrapper, legacy compat or leave as is?

This probably needs to be handled in each writer. The approach isn't scalable in the long-term, but it is what I've been doing so far:

https://github.com/ssec/polar2grid/blob/f19c604d03c0f4fc1452a7ec48a1b484f2349cf1/polar2grid/writers/geotiff.py#L48-L54

And:

https://github.com/ssec/polar2grid/blob/f19c604d03c0f4fc1452a7ec48a1b484f2349cf1/polar2grid/writers/binary.py#L50-L56

The default `None` option matches what we expected in Geo2Grid which is maybe not useful. The other options are for polar2grid. Maybe I should add another level to these dictionaries. One for P2G and one for G2G so the defaults work in more cases.

polar2grid/writers/hdf5.py
```diff
@@ -60,7 +60,14 @@
...
 
 # reader_name -> filename
-DEFAULT_OUTPUT_FILENAMES = {None: "{platform_name}_{sensor}_{start_time:%Y%m%d_%H%M%S}.h5"}
+DEFAULT_OUTPUT_FILENAMES = {
+    None: "{platform_name!u}_{sensor!u}_{p2g_name}_{start_time:%Y%m%d_%H%M%S}_{area.area_id}.tif",
+    "abi_l1b": "{platform_name!u}_{sensor!u}_{observation_type}{scene_abbr}_"
+    "{p2g_name}_{start_time:%Y%m%d_%H%M%S}_{area.area_id}.tif",
+    "viirs_sdr": "{platform_name!l}_{sensor!l}_{p2g_name}_{start_time:%Y%m%d_%H%M%S}_{area.area_id}.tif",
+    "mirs": "{platform_name!l}_{sensor!l}_{p2g_name}_{start_time:%Y%m%d_%H%M%S}_{area.area_id}.tif",
+    "virr_l1b": "{platform_name!l}_{sensor!l}_{p2g_name}_{start_time:%Y%m%d_%H%M%S}_{area.area_id}.tif",
```
djhoese, line 69: Change `.tif` to `.h5`

djhoese: Oh, actually you should remove `p2g_name` too and probably `area id`? Not sure what the old P2G produced.

joleenf: I reverted this back to the single line default.  Looking at old test output on bumi, hdf5 data all have the same default filename convention.

djhoese: Yeah, that's fine. This writer didn't exist for Geo2Grid which is really the only reason there are different naming patterns in these dictionaries anyway.

polar2grid/readers/virr_l1b.py
```diff
@@ -63,112 +64,58 @@
...
+
+    is_polar2grid_reader = True
+
+    def get_default_products(self) -> list[str]:
+        """Get products to load if users hasn't specified any others."""
+        return DEFAULT_PRODUCTS
+
+    def get_all_products(self) -> list[str]:
+        """Get all polar2grid products that could be loaded."""
+        return DEFAULT_PRODUCTS
```
djhoese, line 107: This should also have `ALL_ANGLES` to match line 104 of the old version.

djhoese:
```suggestion
        return ALL_BANDS + ALL_ANGLES + ALL_COMPS
```

PR #367, opened 2021-08-10
Use mask_and_scale in comparison of netcdf

joleenf:
This PR updates the reading of netcdf files to use mask and scale (previously set to false).  xarray.open_dataset replaces netcdf4.Dataset so that mask, scale and fill are handled when read.  

file_type and plot keywords did not seem to be used previously, or something went wrong with my version, this PR updates the use of the --plot and file_type arguments.

the comparison plots were nice, but sometimes hard to know how severe the differences are by just looking at side to side plot.  This adds a nanquantile subtitle to the plots and a plot difference of array1 - array2.

Updates needed for clavrx resampling were found when testing this version of compare, these are added in this PR.

### DIFF
--- a/polar2grid/compare.py
+++ b/polar2grid/compare.py
@@ -126,11 +126,10 @@ def plot_array(array1, array2, cmap="viridis", vmin=None, vmax=None, **kwargs):
     fig.colorbar(img2, ax=ax1[1])
 
     ax2[0].set_title("Difference")
-    inner_fence_low = q[1] - (q[3] - q[1]) * 1.5
-    inner_fence_high = q[3] + (q[3] - q[1]) * 1.5
-    img4 = ax2[0].imshow(array3, cmap=cmap, vmin=inner_fence_low, vmax=inner_fence_high)
+    array3[array3 == 0.0] = np.nan
+    img4 = ax2[0].imshow(array3, cmap=cmap)
     fig.colorbar(img4, ax=ax2[0])
-    fig.delaxes(ax2[1])
+    ax2[1].hist(array3[~np.isnan(array3)], 100)
 
     plt.tight_layout()
     plt.show()
--- a/polar2grid/compare.py
+++ b/polar2grid/compare.py
@@ -117,15 +117,22 @@ def plot_array(array1, array2, cmap="viridis", vmin=None, vmax=None, **kwargs):
     array3 = array1 - array2
     q = [0, 0.25, 0.5, 0.75, 1.0]
     array3_quantiles = np.nanquantile(array3, q)
-    fig.suptitle(array3_quantiles)
+    subtitle = np.array2string(array3_quantiles,  precision=8, separator=',',
+                               suppress_small=True)
+    fig.suptitle(subtitle)
+    img1 = ax1[0].imshow(array1, cmap=cmap, vmin=vmin, vmax=vmax)
+    fig.colorbar(img1, ax=ax1[0])
+    img2 = ax1[1].imshow(array2, cmap=cmap, vmin=vmin, vmax=vmax)
+    fig.colorbar(img2, ax=ax1[1])
 
-    ax1[0].imshow(array1, cmap=cmap, vmin=vmin, vmax=vmax)
-    ax1[1].imshow(array2, cmap=cmap, vmin=vmin, vmax=vmax)
-    diff_max = max(np.nanmax(array3), np.absolute(np.nanmin(array3)))
-    img3 = ax2[0].imshow(array3, cmap='RdBu', vmin=-diff_max, vmax=diff_max)
-    fig.colorbar(img3, ax=ax2[0])
+    ax2[0].set_title("Difference")
+    inner_fence_low = q[1] - (q[3] - q[1]) * 1.5
+    inner_fence_high = q[3] + (q[3] - q[1]) * 1.5
+    img4 = ax2[0].imshow(array3, cmap=cmap, vmin=inner_fence_low, vmax=inner_fence_high)
+    fig.colorbar(img4, ax=ax2[0])
     fig.delaxes(ax2[1])
 
+    plt.tight_layout()
     plt.show()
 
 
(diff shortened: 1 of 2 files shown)


polar2grid/compare.py
```diff
@@ -110,19 +112,33 @@ def plot_array(array1, array2, cmap="viridis", vmin=None, vmax=None, **kwargs):
...
+    fig.suptitle(array3_quantiles)
+
+    ax1[0].imshow(array1, cmap=cmap, vmin=vmin, vmax=vmax)
+    ax1[1].imshow(array2, cmap=cmap, vmin=vmin, vmax=vmax)
+
+    
+    diff_max = max(np.nanmax(array3),np.absolute(np.nanmin(array3)))
+    img3 = ax2[0].imshow(array3, cmap='RdBu', vmin=-diff_max, vmax=diff_max)
+    fig.colorbar(img3, ax=ax2[0])
+    fig.delaxes(ax2[1])
```
djhoese, line 129: Is this specific to the testing you are doing or should this be the new way these plots are done?

polar2grid/compare.py
```diff
@@ -110,19 +112,33 @@ def plot_array(array1, array2, cmap="viridis", vmin=None, vmax=None, **kwargs):
...
 
 def compare_array(array1, array2, plot=False, **kwargs) -> ArrayComparisonResult:
+
     if plot:
         plot_array(array1, array2, **kwargs)
     return isclose_array(array1, array2, **kwargs)
 
 
-def compare_binary(fn1, fn2, shape, dtype, atol=0.0, margin_of_error=0.0, **kwargs) -> list[ArrayComparisonResult]:
+def compare_binary(fn1, fn2, shape, dtype,  atol=0.0, margin_of_error=0.0, **kwargs) -> list[ArrayComparisonResult]:
```
djhoese, line 141:
```suggestion
def compare_binary(fn1, fn2, shape, dtype, atol=0.0, margin_of_error=0.0, **kwargs) -> list[ArrayComparisonResult]:
```

polar2grid/compare.py
```diff
@@ -158,7 +174,7 @@ def _get_geotiff_array(gtiff_fn, dtype=None):
     return arr
 
 
-def compare_awips_netcdf(nc1_name, nc2_name, atol=0.0, margin_of_error=0.0, **kwargs) -> list[ArrayComparisonResult]:
+def compare_awips_netcdf(nc1_name, nc2_name, conovert_nan=None, atol=0.0, margin_of_error=0.0, **kwargs) -> list[ArrayComparisonResult]:
```
djhoese, line 177: Typo and unused?

polar2grid/compare.py
joleenf: I prefer to see a few numbers and a difference plot, but if it causes other issues, then this change can be rejected.

polar2grid/compare.py
```diff
@@ -277,6 +291,7 @@ def compare_files(self, file1, file2, file_type=None, **kwargs):
             # guess based on file extension
             ext = os.path.splitext(file1)[-1]
             file_type = file_ext_to_compare_func.get(ext)
+            file_type = file_ext_to_compare_func.get(ext)
```
djhoese, line 294: This is a duplicate of the line above it, right?

joleenf: Yes, not sure how that got duplicated.

polar2grid/compare.py
```diff
@@ -302,6 +317,7 @@ def compare_dirs(self, dir1, dir2, **kwargs) -> list[FileComparisonResults]:
         return results
 
     def compare(self, input1, input2, **kwargs) -> list[FileComparisonResults]:
+        kwargs.update({"plot": self.create_plot})
```
djhoese, line 320: You are right that this is missing, but I think it might be cleaner if you move this to the `compare_files` method similar to how `self.margin_of_error` is used.

polar2grid/compare.py
joleenf: Oh.  I missed that, this was something I had done before changing to xarray and noticing how the isclose comparison used equal_nan.  I will remove this.

joleenf: Not only that, it was a typo (spelled incorrectly).

polar2grid/compare.py
```diff
@@ -512,7 +528,8 @@ def main(argv=sys.argv[1:]):
 
     levels = [logging.ERROR, logging.WARN, logging.INFO, logging.DEBUG]
     logging.basicConfig(level=levels[min(3, args.verbosity)])
-    compare_kwargs = {"shape": tuple(args.shape), "dtype": args.dtype, "variables": args.variables}
+    compare_kwargs = {"shape": tuple(args.shape), "dtype": args.dtype, "variables": args.variables,
+                      "file_type": args.file_type}
```
djhoese, line 532: Good catch. Could you verify that if this is not provided that `args.file_type` is `None` (versus `[]`). If it is `[]` then it won't get detected later on when we do `file_type is None`.

polar2grid/compare.py
djhoese: No problem. Appreciated addition. I just wanted to make sure this wasn't left over from debugging. Could you remove the extra blank line above `diff_max = `?

polar2grid/compare.py
```diff
@@ -101,7 +103,7 @@ def isclose_array(array1, array2, atol=0.0, rtol=0.0, margin_of_error=0.0, **kwa
     LOG.info("%d pixels out of %d pixels are different" % (diff_pixels, total_pixels))
     return ArrayComparisonResult(True, diff_pixels, total_pixels, False)
 
-
+
```
djhoese, line 106:
```suggestion

```

polar2grid/compare.py
```diff
@@ -110,19 +112,33 @@ def plot_array(array1, array2, cmap="viridis", vmin=None, vmax=None, **kwargs):
...
+    fig, (ax1, ax2) = plt.subplots(2, 2)
+    array3 = (array1 - array2)
+    q=[0,0.25,0.5, 0.75,1.0]
+    array3_quantiles = np.nanquantile(array3, q)
+    fig.suptitle(array3_quantiles)
+
+    ax1[0].imshow(array1, cmap=cmap, vmin=vmin, vmax=vmax)
+    ax1[1].imshow(array2, cmap=cmap, vmin=vmin, vmax=vmax)
+
+
```
djhoese, line 125:
```suggestion

```

polar2grid/compare.py
joleenf: Okay.  I will work to update this.

djhoese (approved):
Thank you!

Issue #303, opened 2021-02-01
Add caching option to add_coastlines


djhoese:
Pycoast, the library we use for the add_coastlines script, has a function that can cache the coastlines that are drawn. See https://pycoast.readthedocs.io/en/latest/api/pycoast.html#pycoast.cw_base.ContourWriterBase.add_overlay_from_dict for details. If we refactored the add_coastlines script to put all available options in a dictionary and then pass them to this function (instead of individual calls). This should have a huge performance impact on the script when used on things like Geo2Grid output. Polar2Grid will not really benefit unless it is used on a static grid (one that doesn't change from swath to swath).

jcovington16:
So just to reiterate so I understand better... You want to cache all available options into a dictionary of its own and then pass that dictionary into the function so it can call all the arguments in that dictionary?

jcovington16:
Seeing all the different arguments you have listed are there some cases where the user won't need every option available? Throwing these arguments into a dictionary for them to be passed in automatically sounds like it will be easier for the use... but what if there are options I want to not use in a particular use case. I guess what I'm asking is do you still want the user to use these arguments, but index through the dictionary to pass them into the function? Just trying to understand better so I can help.

djhoese:
From the user's perspective the command line arguments in `add_coastlines` will not change. The reason things have to be thrown into a dictionary is because the method in pycoast (`add_overlay_from_dict`) takes a dictionary. The dictionary will only need the options added for the ones that the user specifies as the defaults (in `add_overlay_from_dict`) are to not perform those actions if they are not present. It is essentially offloading the work from the add_coastlines command line arguments to the `add_overlay_from_dict` function.

jcovington16:
So from reading the pycoast documentation which I will paste in here for you describes taking in a configuration file as well if that is an option. Two things can happen. I can create a script to have the user throw in the options they want for each argument that requires it, throw it into a configuration file and have that file passed in or throw it in a dictionary like you are requesting and pass in that dictionary. Either way there needs to be something that will gather the options the user wants so that it can be stored in the dictionary since add_coastlines will not change. 

Also to clarify you want to include the add_overaly_from_dict() inside of the add_coastline script to pass into your main()

[https://pycoast.readthedocs.io/_/downloads/en/latest/pdf/](url)

jcovington16:
My apologies. I've missed spoke was not on the same page. I totally understand what needs to happen. I just forked the repo and will get started on it if that's cool?

[...]

PR #368, opened 2021-08-10
Add --extra-config-path glue option and add_coastlines optional caching

djhoese:
Currently it is only possible to have per-run config changes if the user manually changes the directory built into the polar2grid bundle (etc). This new `--extra-config-path` flag lets you specify one or more additional paths (can be specified more than once) with Satpy YAML configuration files.

This PR includes updating the AMSR2 example which used --rescale-configs so it now uses this new flag with a new enhancement configuration file.


djhoese:
With my latest commits this actually implements the caching for add_coastlines.sh too (closes #303).

Issue #380, opened 2021-09-08
Ongoing documentation issues for Polar2Grid 3.0


djhoese:
- [x] Change crefl reader examples to "Product Explanation" to describe true_color, false_color, and fog products. Make sure to add "crefl" rayleigh correction section to image processing techniques page.
- [x] Add information on YAML grid/area configs
- [x] Add "--to-yaml" option to p2g_grid_helper.sh
- [x] Add grid_to_yaml.sh helper script
- [x] Convert "add_colormap.sh" documentation about "rescale" to enhancement example
- [x] Review "Python Proj" documentation and see if custom grid configuration docs can be updated to not reference it.
- [x] Move "rescale" documentation to Satpy with nice math formulas (possibly need to add extensions to render them)
- [x] Rewrite VIIRS crefl verification example
- [x] Rewrite MODIS crefl verification example
- [x] Rewrite VIIRS example
- [x] Rewrite MODIS example
- [x] Rewrite ACSPO example (`--extra-config-path` already added and pointing to new YAML file)
- [x] Rewrite AMSR2 example
- [x] Convert builtin grids to new `areas.yaml` format
- [x] Convert example grids file in `swbundle/` to new YAML format
- [x] Add geo2grid documentation generation to CI
- [x] Review misc_recipes (can they go somewhere else)
- [x] Update misc_recipes to not talk about GDAL binaries being in `./bin/` (they are in the vendor runtime directory).
- [x] Debug why `.. only:: html` section is added to PDF

New items as of 2021-09-15:

- [x] Add documentation and links for customizing Polar2Grid/Satpy configuration directories (`--extra-config-path`)

djhoese:
- [ ] Fix helper script not specifying units for degree center points

kathys:
These have all been completed.

PR #383, opened 2021-09-21
Switch grid helper script to produce YAML instead of legacy format

djhoese:
Also includes a rewrite of the grid configuration documentation and automatically generating the list of grids in the documentation.


djhoese:
Related to #380

PR #424, opened 2022-03-02
Fix warning produced by --dnb-saturation-correction flag

### LINKED ISSUE #420: Using --dnb-saturation-correction flag results in all black VIIRS DNB images
We often use the --dnb-saturation-correction option when creating nighttime DNB images to make sure the images are not overwhelmed by things like city lights and the aurora.  When using this option while executing the most recent version of P2G Version 3.0 (polar2grid-swbundle-20211219-125805) a warning message is printed and the resultant output is an all black image.

**execution:**  polar2grid.sh -r viirs_sdr -w geotiff -p dynamic_dnb --dnb-saturation-correction -f ../input_sdr_night_2022

**execution directory:** bumi:/data/users/kathys/version_3_0_timing/tmp

The execution directory contains the output geotiff files that result when using the option and not using the option.  Data files and log files are in the same directory.

djhoese:
Closes #420

Currently this only implements the fix for the warning message. Summary of the problem is that the flag was causing the list of products to load to be `["dynamic_dnb", "dynamic_dnb"]` and then when P2G was looking up what name this might refer to in Satpy-land it was seeing it already had something looked up. Simple fix, but I made it in two places to be sure.

I still need to look into the black images.

### DIFF
--- a/polar2grid/glue.py
+++ b/polar2grid/glue.py
@@ -856,10 +856,10 @@ def main(argv=sys.argv[1:]):
         _print_list_products(reader_info, p2g_only=not args.list_products_all)
         return 0
 
-    load_args["products"] = reader_info.get_satpy_products_to_load()
-    if not load_args["products"]:
+    load_args.pop("products")
+    products = reader_info.get_satpy_products_to_load()
+    if not products:
         return -1
-    products = load_args.pop("products")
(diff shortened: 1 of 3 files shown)


djhoese:
Fix for the black image has been merged in Satpy so this can be merged now too and the issue closed.

PR #426, opened 2022-03-04
Add AMI L1B (ami_l1b) reader

djhoese:
Adds new GK-2A AMI L1B reader.

```
$ geo2grid.sh -r ami_l1b -w geotiff -f /data/satellite/ami/20190930_0300 --list-products
INFO     : Sorting and reading input files...
INFO     : Loading product metadata from files...
INFO     : Using default product list: ['VI004', 'VI005', 'VI006', 'VI008', 'NR013', 'NR016', 'SW038', 'WV063', 'WV069', 'WV073', 'IR087', 'IR096', 'IR105', 'IR112', 'IR123', 'IR133', 'true_color', 'natural_color']
IR087
IR096
IR105
IR112
IR123
IR133
NR013
NR016
SW038
VI004
VI005
VI006
VI008
WV063
WV069
WV073
airmass
ash
natural_color
true_color
```

PR #428, opened 2022-03-05
Add AGRI L1 (agri_l1) reader

djhoese:
This PR is based on #426 which should be merged first. This adds the `agri_l1` reader...or at least an initial version.


djhoese:
This is going to take some work. Something is going on with this reader. I can't generate a true_color in 12+ minutes.

Issue #422, opened 2022-03-01
Polar2Grid v3.0 MODIS geotiff differences from version 2.3 output


kathys:
I made a montage from all of the default output GeoTIFF band images MODIS from a 3 granule overpass over China, and there are two distinct differences that can be seen:

1).  P2G Version 3.0 shows lots of black in the Band 2 image.  This band saturates alot, and I believe there is code to replace that saturated region to make a better image in P2G V2.3.
Band 2 images:
![aqua_modis_vis02_20210318_060500_wgs84_fit_v2](https://user-images.githubusercontent.com/1848019/156254426-68c6e47f-8a3a-4219-a8e5-1aa1a6a84a16.jpg)

![aqua_modis_vis02_20210318_060500_wgs84_fit_v3](https://user-images.githubusercontent.com/1848019/156254438-f9c61e22-3215-4b86-8244-ae4ecad55200.jpg)

2).  The Band 21 images also are enhanced a differently.   These differences are bit more subtle, see below

![aqua_modis_bt21_20210318_060500_wgs84_fit_v2](https://user-images.githubusercontent.com/1848019/156254863-2f7ca36d-4be1-42d7-b6ba-e2783fab24a2.jpg)

![aqua_modis_bt21_20210318_060500_wgs84_fit_v3](https://user-images.githubusercontent.com/1848019/156255046-6864ea69-9ce0-4c89-a585-fc9fff7f6e7a.jpg)

djhoese:
Do you have log files for the band 21 image?

djhoese:
Putting this here for my own reference, here is the old values we used:

https://github.com/ssec/polar2grid/blob/ab3cbe313c7fd7f78a2c71cb135d647da61d688c/polar2grid/modis/modis_guidebook.py#L368-L372

And where it was used:

https://github.com/ssec/polar2grid/blob/ab3cbe313c7fd7f78a2c71cb135d647da61d688c/polar2grid/modis/modis_guidebook.py#L442-L444

The satpy code does not handle these saturation/can't aggregate values at all:

https://github.com/pytroll/satpy/blob/40148465a8544ee91646aa2be745cceca66f281f/satpy/readers/modis_l1b.py#L123-L125

kathys:
Log file: bumi:/data/users/kathys/version_3_0_timing/output_modis_v3_0/modis_l1b_geotiff_20210318_060500.log

This implementation was for all default MODIS bands, including Band 21.

djhoese:
[...]

PR #435, opened 2022-03-11
Add command line option to mask saturated modis_l1b band 2 pixels

djhoese:
This has to be in the command line so it automatically gets passed to Satpy to turn off masking of saturated pixels. Closes the saturated portion of #422. I will try to find a pretty way to hide it.


djhoese:
Ok so I just tested this branch now that the Satpy PR is merged. If I add the `--mask-saturated` flag to turn this OFF then I get:

image

Without that flag (the default):

image

Also note that this flag is not visible to users in the usage text.

PR #459, opened 2022-04-18
Fix bundle initialization to use embedded openssl

wroberts4:
In order to activate the environment in swbundle/env.sh, METADATA_CHECKSUM  is set using openssl. However, since the environment is not activate yet, it cannot find openssl on all systems (rockylinux:8 docker image). This merge points to the packaged environment's openssl directly.


djhoese:
It was noted by Scott Nolin on slack that sha256sum and sha512sum are part of the rocky 8 "coreutils" packages so they should be widely available. I think you had noted @wroberts4 that they were in your docker image. I see these on ubuntu:latest too. This script didn't need to run on Windows or MacOS so maybe these other tools would be good to depend on. :fearful: 

Maybe we just leave this like it is until we find a situation where this fails. Thanks @wroberts4!

PR #461, opened 2022-04-20
Add tests for main and compare script

djhoese:
Noticed that the compare script and the `__main__` script are completely uncovered. This PR adds unit tests for them for future modifications/features.


djhoese:
This PR also removes the legacy AWIPS NetCDF3 format from the compare script as a possibility.

Issue #471, opened 2022-05-12
netcdf or hdf writer


joleenf:
I am thinking I just missed an issue request already open for this.  Are there plans to facilitate netcdf or hdf writing of files with geo2grid?  Currently planning on using satpy unless there is a geo2grid work around.

djhoese:
So there is the `hdf5` writer, but this is mainly meant for backwards compatibility with older Polar2Grid versions. It is not in Satpy and only available in P2G/G2G.

Technically the `cf` writer should be usable right now, but since it isn't configured/added in P2G/G2G you won't be able to customize any of the parameters. I'm not against added it.

joleenf:
Hi Dave,

"cf" is not recognized as a known writer argument.  That writer is in satpy, would some sort of interface in polar2grid needs to be written in order for that to work?

djhoese:
Darn. I had hoped that G2G/P2G would let you pass any satpy writer and it would "just work". So in this case, yes, a wrapper would need to be written similar to the geotiff writer. We could maybe update P2G/G2G to allow any random writer that exists in Satpy, but in this case you'll want to add command line options to map to kwargs in the satpy writer anyway.

djhoese:
I think this is close-able now that #476 was merged a long time ago. Let me know if we need to reopen it. Or we can create a new issue.

PR #476, opened 2022-06-07
Add CF-complaint NetCDF writer (cf)

joleenf:
Meant to resolve issue #471.  

1. The DEFAULT_OUTPUT_FILENAMES are wrong if an alternate engine is used.  What is the best way to deal with this in polar2grid...no default?
2. The boolean options are reversed for glue.  Most use a flag that negate the satpy flag.  Is there a better way to deal with this.
3. Command line dictionaries get pretty ugly in the glue.py call.  Is there a preferred way to deal with this in polar2grid?
ex.). python ~/code/polar2grid/polar2grid/glue.py -r clavrx -w cf -f /Users/joleenf/data/p2g/clavrx/meso/clavrx_OR_ABI-L1b-RadC-M6C01_G16_s202212419011712.level2.nc -g WGS84_MESO --grid-configs mygrid.conf -p cld_press_acha cld_temp_acha --groups "{\ \"WGS84_MESO\\": [\\"cld_press_acha\\", \\"cld_temp_acha\\"]}" --method nearest -vvv

### DIFF
--- a/polar2grid/writers/cf.py
+++ b/polar2grid/writers/cf.py
@@ -35,8 +35,8 @@ different grids, the CF compliant workaround is to save the datasets to separate
 
 
 """
-import logging
 import json
+import logging
 
 LOG = logging.getLogger(__name__)
 
@@ -52,7 +52,6 @@ DEFAULT_OUTPUT_FILENAMES = {
 
 
 def add_writer_argument_groups(parser, group=None):
-
     if group is None:
         group = parser.add_argument_group(title="cf Writer")
     group.add_argument(
--- a/polar2grid/writers/cf.py
+++ b/polar2grid/writers/cf.py
@@ -45,7 +45,7 @@ DEFAULT_OUTPUT_FILENAMES = {
     "polar2grid": {
         None: "{platform_name}_{sensor}_{start_time:%Y%m%d_%H%M%S}.nc",
     },
-    "geo2grid": {
+        "geo2grid": {
         None: "{platform_name}_{sensor}_{start_time:%Y%m%d_%H%M%S}.nc",
     },
 }
--- a/polar2grid/writers/cf.py
+++ b/polar2grid/writers/cf.py
@@ -38,13 +38,11 @@ different grids, the CF compliant workaround is to save the datasets to separate
 import logging
 import json
 
-from polar2grid.core.script_utils import BooleanOptionalAction
-
 LOG = logging.getLogger(__name__)
 
 # reader_name -> filename
 DEFAULT_OUTPUT_FILENAMES = {
-        "polar2grid": {
+    "polar2grid": {
         None: "{platform_name}_{sensor}_{start_time:%Y%m%d_%H%M%S}.nc",
     },
     "geo2grid": {
@@ -54,7 +52,6 @@ DEFAULT_OUTPUT_FILENAMES = {
 
 
 def add_writer_argument_groups(parser, group=None):
-
     if group is None:
         group = parser.add_argument_group(title="cf Writer")
     group.add_argument(
@@ -80,8 +77,8 @@ def add_writer_argument_groups(parser, group=None):
         "--engine",
         default="netcdf4",
         help="Module to be used for writing netCDF files. Follows xarray's"
-              ":meth:`~xarray.Dataset.to_netcdf` engine choices with a"
-              "preference for 'netcdf4'.",
+             ":meth:`~xarray.Dataset.to_netcdf` engine choices with a"
+             "preference for 'netcdf4'.",
     )
     group.add_argument(
         "--epoch",
--- a/polar2grid/writers/cf.py
+++ b/polar2grid/writers/cf.py
@@ -27,12 +27,17 @@
 #     1225 West Dayton Street
 #     Madison, WI  53706
 #     david.hoese@ssec.wisc.edu
-"""The cf writer puts gridded image data into a `CF-compliant` netCDF file.
+r"""The cf writer puts gridded image data into a `CF-compliant` netCDF file.
 
 All datasets to be saved must have the same projection coordinates ``x`` and ``y``. If a scene holds datasets with
 different grids, the CF compliant workaround is to save the datasets to separate files.
 
+Group key and list values  must use escapes before quoting.
 
+Example:
+  >>>  python glue.py -r clavrx -w cf -f clavrx_OR_ABI-L1b-RadC-M6C01_G16_s202212419011712.level2.nc \
+       -vvv -g WGS84_MESO --grid-configs mygrid.conf -p cld_press_acha cld_temp_acha --method nearest \
+       --groups "{\"WGS84_MESO\": [\"cld_press_acha\", \"cld_temp_acha\"]}"
 
 """
 import json
--- a/polar2grid/writers/cf.py
+++ b/polar2grid/writers/cf.py
@@ -38,11 +38,13 @@ different grids, the CF compliant workaround is to save the datasets to separate
 import logging
 import json
 
+from polar2grid.core.script_utils import BooleanOptionalAction
+
 LOG = logging.getLogger(__name__)
 
 # reader_name -> filename
 DEFAULT_OUTPUT_FILENAMES = {
-    "polar2grid": {
+        "polar2grid": {
         None: "{platform_name}_{sensor}_{start_time:%Y%m%d_%H%M%S}.nc",
     },
         "geo2grid": {
@@ -52,6 +54,7 @@ DEFAULT_OUTPUT_FILENAMES = {
 
 
 def add_writer_argument_groups(parser, group=None):
+
     if group is None:
         group = parser.add_argument_group(title="cf Writer")
     group.add_argument(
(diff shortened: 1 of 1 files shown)


djhoese:
FYI the `main` branch should now have the fixes necessary to make the "Build Documentation" job succeed.

joleenf:
:thumbsup:  It worked!  I fixed the escape characters in the comment above.  (I needed to escape the escape characters).

joleenf:
Strange.  All checks passed here, but I received an email message that the build documentation failed in 7 minutes this morning.

djhoese:
That might have been on your own fork here: https://github.com/joleenf/polar2grid/runs/6830437586?check_suite_focus=true

Not sure what's going on exactly.

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python3
+# encoding: utf-8
+# Copyright (C) 2018 Space Science and Engineering Center (SSEC),
```
djhoese, line 3:
```suggestion
# Copyright (C) 2022 Space Science and Engineering Center (SSEC),
```

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+# satellite observation data, remaps it, and writes it to a file format for
+# input into another program.
+# Documentation: http://www.ssec.wisc.edu/software/polar2grid/
+#
+#     Written by David Hoese    March 2016
+#     University of Wisconsin-Madison
+#     Space Science and Engineering Center
+#     1225 West Dayton Street
+#     Madison, WI  53706
+#     david.hoese@ssec.wisc.edu
```
djhoese, line 29: I've started removing this bit from the header. It just feels weird to include it everywhere.

```suggestion
```

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+"""The cf writer puts gridded image data into a `CF-compliant` netCDF file.
```
djhoese, line 30:
```suggestion
"""The CF writer puts gridded image data into a `CF-compliant` NetCDF file.
```

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        dest="header_attrs",
+        type=json.loads,
+        help="Global attributes to be included.",
+    )
+
+    group.add_argument(
+        "--engine",
+        default="netcdf4",
+        help="Module to be used for writing netCDF files. Follows xarray's"
+        ":meth:`~xarray.Dataset.to_netcdf` engine choices with a"
```
djhoese, line 78: I don't think we can use intersphinx references here like `:meth:` as the actual command line output will not be converted to a URL (that'd be really cool if it did). You might need to either include a URL to the page you're talking about or only mention xarray and don't link to anything.

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        "preference for 'netcdf4'.",
+    )
+    group.add_argument(
+        "--epoch",
+        help="Reference time for encoding of time coordinates",
```
djhoese, line 83: I know this is taken from the satpy documentation for this parameter, but maybe this should have "units" in the name somewhere since as far as I can tell that's what it is used for.

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+    group.add_argument(
+        "--groups",
+        dest="groups",
+        type=json.loads,
+        help="Group datasets according to the given assignment: `{'group_name': ['dataset1', 'dataset2', ...]}`.",
+    )
+
+    group.add_argument(
+        "--header_attrs",
```
djhoese, line 68: All command line options should have `_` replaced with `-`. Ex: `--header-attrs`. `argparse` will automatically rename the destination with underscores so you don't need to specify `dest`, it will be `args.header_attrs`.

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        ":meth:`~xarray.Dataset.to_netcdf` engine choices with a"
+        "preference for 'netcdf4'.",
+    )
+    group.add_argument(
+        "--epoch",
+        help="Reference time for encoding of time coordinates",
+    )
+
+    group.add_argument(
+        "--exclude_attrs",
```
djhoese, line 87: Hyphen

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        nargs="+",
+        help="List of dataset attributes to be excluded",
+    )
+    group.add_argument(
+        "--no_lonlats",
```
djhoese, line 92: Hyphen

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        dest="include_lonlats",
+        action="store_false",
```
djhoese, line 94: Try using:

```
from polar2grid.core.script_utils import BooleanOptionalAction

    group.add_argument(
        "--tiled", action=BooleanOptionalAction, default=True, help="Tile geotiffs internally"
    )
```

This means there is a `--tiled` and `--no-tiled` option available. The `default=True` means that `--tiled` is the default. This gets around the odd behavior of the usage saying true or false is default when the `dest` is the opposite of the flag name (`no-lonlats` versus `include-lonlats`). So I think with this usage you could name this flag `--include-lonlats` and default it to `True` so someone could then specify `--no-include-lonlats` if they wanted.

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        help="Don't include latitude and longitude coordinates.",
+    )
+
+    group.add_argument("--not_pretty", dest="pretty", action="store_true", help="Modify coordinate names")
```
djhoese, line 98: Same Boolean action here maybe. And hyphen.

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+
+    group.add_argument(
+        "--no_include_orig_name",
```
djhoese, line 101: boolean  and hyphen

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        dest="include_orig_name",
+        action="store_false",
+        help="Do not include the original dataset name as a variable attribute in the final netcdf.",
+    )
+
+    group.add_argument(
+        "--flatten_attrs",
```
djhoese, line 108: hyphen

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        action="store_true",
+        help="If invoked, flatten dict-type attributes",
+    )
+
+    group.add_argument(
+        "--numeric_name_prefix",
```
djhoese, line 114: hyphen

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,119 @@
...
+        default="CHANNEL_",
+        help="Prefix added to each variable. For name starting with a digit.Use '' or None to leave this out..",
```
djhoese, line 116:
```suggestion
        help="Prefix added to each variable. For names starting with a digit, use '' or None to leave this out.",
```

I think you could set `nargs="?"` which would allow this to be customized with an argument or left blank to be turned off. Otherwise you may need to have a special `type` handler to do the handling of `None` as a string to `None` the python object.

djhoese (changes requested):
Lots of comments and suggestions. Thanks for putting this together. It'll be nice to have. A couple other things:

1. It'd be nice to have some tests, but I see we don't have any tests for command-line only writers anyway...at least not specific to the writer.
2. I'm not sure I understood your first question about the default output filenames being wrong depending on the engine used?

joleenf:
> 1. It'd be nice to have some tests, but I see we don't have any tests for command-line only writers anyway...at least not specific to the writer.

So, no tests for now?  Or start thinking about creating tests?  

> 2. I'm not sure I understood your first question about the default output filenames being wrong depending on the engine used?

My mistake...The cf writer should only be writing NetCDF.

polar2grid/writers/cf.py
joleenf: You mean, --lon-lats by same boolean action?

joleenf:
```python
group.add_argument(
        "--include-lonlats",
        dest="include_lonlats",
        action="store_true",
        help="Include latitude and longitude coordinates.",
    )
```
? So that the actual flag used has a hyphen?

djhoese: Yes the flag has the hyphen and remove `dest`.

polar2grid/writers/cf.py
```diff
@@ -0,0 +1,110 @@
...
+        "--engine",
+        default="netcdf4",
+        help="Engine for writing NetCDF file \
+             (options: '"
+        "netcdf4"
+        "', '"
+        "scipy"
+        "', '"
+        "h5netcdf"
+        "').",
```
djhoese, line 75: Does this show up in the helm message properly?

djhoese (approved):
Besides the question about the usage formatting, if this works for you then it looks good to me.

Issue #423, opened 2022-03-01
MODIS overpass version 2.3 version 3.0 processing speeds


kathys:
I wrote scripts to time the creation of all P2G MODIS GeoTIFF default bands for a 15 minutes pass using the default WGS84 dynamic grid.  For Version 2.3 P2G I added the times together that it took create the default images for crefl (true and false color) and for the vis/ir bands.  For P2G version 3.0, I used 4 workers, which I think is the default.  Here are the results:

P2G Version 2.3:  8m22s
P2G Version 3.0:  15m33s

It took almost twice as long using more CPU's to create the images using P2G v3.0 than it did with V2.3 on the machine bumi.  I cannot imagine releasing software until this is improved.  Dynamic grids are the way the software is most used by Liam and I in SSEC, and I suspect most used by the community too.

djhoese:
Could you give me the exact commands run and individual timings for each command if you have them (if not that's fine).

kathys:
Version 3.0:

/data/users/kathys/polar2grid-swbundle-20211219-125805/bin/polar2grid.sh -r modis_l1b -w geotiff --num-workers 4 --fill-value 0 -f /data/users/kathys/version_3_0_timing/input_modis_china
INFO     : Sorting and reading input files...
INFO     : Loading product metadata from files...
INFO     : Using default product list: ['vis01', 'vis02', 'vis03', 'vis04', 'vis05', 'vis06', 'vis07', 'vis26', 'bt20', 'bt21', 'bt22', 'bt23', 'bt24', 'bt25', 'bt27', 'bt28', 'bt29', 'bt30', 'bt31', 'bt32', 'bt33', 'bt34', 'bt35', 'bt36', 'true_color', 'false_color', 'fog']
INFO     : Running day coverage filtering...
INFO     : Running night coverage filtering...
INFO     : Computing dynamic grid parameters...
INFO     : Checking products for sufficient output grid coverage (grid: 'wgs84_fit')...
INFO     : Resampling to 'wgs84_fit' using 'ewa' resampling...
INFO     : Computing products and saving data to writers...
INFO     : SUCCESS

real    15m33.582s
user    21m7.667s
sys     8m54.846s

Version 2.3

/data/users/kathys/polar2grid_v_2_3/bin/polar2grid.sh crefl gtiff -f /data/users/kathys/version_3_0_timing/input_modis_china
INFO     : Initializing reader...
INFO     : Could not find any existing crefl output will use MODIS SDRs to create some
[...]

PR #483, opened 2022-06-17
Fix persisting geolocations not actually persisting

djhoese:
In an effort to improve #423, I noticed that the swath persisting logic implemented in #472 wasn't actually holding on to the persisted data. It was a simple logic bug, but an important one. It meant that the lon/lats were being persisted (computed) but then immediately discarded and never used. So it basically did each geolocation load and interpolation at least one extra time more than it had to.

### DIFF
--- a/polar2grid/filters/day_night.py
+++ b/polar2grid/filters/day_night.py
@@ -42,7 +42,7 @@ logger = logging.getLogger(__name__)
 
 
 @cache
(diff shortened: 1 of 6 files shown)


djhoese:
On bumi this seems to bring processing of vis01/vis02/bt20 from ~3m10s to ~1m40s. However, v2.3 still somehow does 1m25s.

PR #501, opened 2022-08-22
Force ACSPO resampling options all to 40:1

### LINKED ISSUE #498: ACSPO EWA Resampling parameters should be set to 40:1 for all instruments.
After looking closely at the Version 2.3 versus 3.0 ACSPO SST images, I think the originals look very good.  V3.0 is using 10:1 for MODIS, which results in holes at the edges of the scan, and we have already demonstrated the improved detail in the 40:1 for VIIRS, so I think the original V2.3 values will work best for MODIS, VIIRS and AVHRR.

djhoese:
Closes #498

### DIFF
--- a/etc/resampling.yaml
+++ b/etc/resampling.yaml
@@ -215,5 +215,5 @@ resampling:
     sensor: avhrr
     resampler: ewa
     kwargs:
-      weight_delta_max: 10.0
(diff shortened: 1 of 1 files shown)


djhoese:
Needs https://github.com/pytroll/satpy/pull/2190 to pass probably.

Issue #503, opened 2022-08-24
add_coastline.sh script not placing labels correctly


kathys:
Using the same `add_coastline.sh`script that I have used in the past on an SST image, results in the colorbar temperature labels being incorrectly displayed.  The range of values that I am using are 275-305K.  The only temperature that is labelled is 275 and it is located in the wrong location along the colorbar. 

The data is located here:  `bumi:/data/users/kathys/test_data/acspo/example_for_documentation/sst_file`

The command I used:
add_coastlines.sh noaa20_viirs_sst_20220810_184327_great_lakes.tif  --add-colorbar --colorbar-text-color="white" --colorbar-units="°K" --colorbar-align bottom --colorbar-title="VIIRS ACSPO SST 10 August 2022 18:43 UTC" --colorbar-text-size 20 --colorbar-height=35

![Image](https://user-images.githubusercontent.com/1848019/186473656-af36d701-4864-44a5-bf36-90b0db09a17e.png)

djhoese:
What are the different files in this directory? What's the difference between `orig` and `rescaled` and `rescaled_wcolor` and `wcolor` and the one with no additional suffix?

kathys:
These are files created as part of the steps involved in the process of creating a color enhanced, customized grid and scaled image for the doc section on ACSPO SSTs.  I did not explain very well about the image I am using to add the colorbar.  The image that I am am applying the add_coastlines.sh to, is the `noaa20_viirs_sst_20220810_184327_great_lakes.tif` which is just a copy of the `noaa20_viirs_sst_20220810_184327_great_lakes_rescaled_wcolor.tif`.  So the image was created using a customized grid and customized scaling with a colortable added.

djhoese:
Should be closed by #507

PR #507, opened 2022-08-29
Fix add_coastlines.py using the wrong colorbar limits when geotiff contains a colormap

djhoese:
This PR fixes an issue that has existed since changes were made to colormap'ing in Satpy/Polar2Grid. There was a change made in trollimage that added scaling and offset information and eventually included a new `colormap` tag with a CSV version of the colormap that was applied. If the colormap is added with `add_colormap.sh` (what I would consider a "legacy" way of doing this), then that attribute is not available. In both of these cases where the geotiff has a colormap built in the add_coastlines script thought scale factor and offset were based on a 0 to 1 scale. This was the original way that trollimage was going to work with these attributes but was changed last minute and apparently I never updated polar2grid to reflect this.

### DIFF
--- a/polar2grid/add_coastlines.py
+++ b/polar2grid/add_coastlines.py
@@ -41,6 +41,7 @@ from aggdraw import Font
 from PIL import Image, ImageFont
 from pkg_resources import resource_filename as get_resource_filename
 from pycoast import ContourWriterAGG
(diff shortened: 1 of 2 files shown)


djhoese:
Closes #503

PR #522, opened 2022-09-14
Fix backwards compatibility for --output-filename in more writers

djhoese:
This PR adds some backwards compatibility for people who used the old (and removed) `--filename-pattern` command line flag. This flag has been renamed `--output-filename` to be more clear about what it is, but now also uses Satpy-based metadata. This PR adds an extra step in this output naming by renaming old filename-pattern elements (ex. `grid_name`) to map to the new naming (ex. `area.area_id`). This mostly already existed thanks to @joleenf, but this PR adds it for more writers.


djhoese:
@kathys I think I'd like to add a log message to this that logs a warning if users use the old names for things. Thoughts?

kathys:
I think it is a great idea to provide a warning to users.  It makes it easy for them to diagnose problems if the file names are different from what they expect.

PR #543, opened 2022-10-14
Deprecate --colorbar-vertical in add_coastlines

### LINKED ISSUE #439: --colorbar-align in geo2grid documentation
If you specify "--colorbar-align right" but don't also specify --colorbar-vertical, the -align request is ignored.  The choices for align are top, bottom, left, right.  Horizontal or Vertical is implied in those choices.  Can --colorbar-vertical be removed and have everything be controlled by --colorbar-align?

djhoese:
Closes #439 

As described in the related issue, there is no need for --colorbar-vertical to exist as it can be inferred from the position of the colorbar (top, bottom, left, right). This PR removes the usage of it but doesn't remove the flag so it doesn't completely break people's scripts.

This PR also re-enables the default colorbar to be a grayscale colormap. This could be a problem if users think their non-linearly scaled data should have a linear colormap, but I think that's up to the user. @kathys what do you think: If I did a square root enhancement on VIIRS I01 and then ran add_coastlines.sh I'd be get a black->white colormap with the tick labels being 0 to 255.

### DIFF
--- a/build_environment.yml
+++ b/build_environment.yml
@@ -20,7 +20,7 @@ dependencies:
   - pillow
   - pip
   - pooch
-  - pycoast>=1.5.0
+  - pycoast>=1.6.0
   - pydecorate>=0.3.3
   - pyhdf
   - pykdtree
--- a/polar2grid/add_coastlines.py
+++ b/polar2grid/add_coastlines.py
@@ -51,21 +51,6 @@ LOG = logging.getLogger(__name__)
 PYCOAST_DIR = os.environ.get("GSHHS_DATA_ROOT")
 
 
-def _convert_table_to_cmap_or_default_bw(band_dtype, band_ct, band_count):
-    max_val = np.iinfo(band_dtype).max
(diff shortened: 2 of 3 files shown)


kathys:
@djhoese   I guess my first question is if this a change from the way this worked in V2.3?

djhoese:
I *think* this is more of a bug fix from v2.3. For my own future reference there is this line of code:

https://github.com/ssec/polar2grid/blob/p2g-v2.3.0b0/polar2grid/add_coastlines.py#L79

I have not actually run the code but this says to me if the geotiff doesn't have a colormap on it then it tries to generate one. The one it generates will be all black due to converting floating point numbers to integers. So effectively a colorbar that is all black. Let me try running some stuff with v2.3...

djhoese:
Yeah v2.3 does an all black colorbar. Maybe I just stick with that even though I'm not sure if it is more right or more wrong than a grayscale colorbar.

PR #568, opened 2022-11-20
Fix VIIRS Active Fires default radius_of_influence and boundary NaN handling

### LINKED ISSUE #561: Problems with handling of VIIRS aggregated files
I am running a test parallel processing direct broadcast system that uses the P2G Version 3.0 (Nov. 4 tarball).  I can directly compare the output products that are made from the operational system that uses P2G V2.3.  My parallel system uses aggregated VIIRS granules, which is what Liam prefers.  I am seeing differences in the handling of the aggregated versus non-aggregated granules with P2G V3.0.  Using the aggregated granules, I sometimes see these warning messages in the log files:
`WARNING  : Geolocation data contains invalid bounding values. Computing entire array to get bounds instead.`  
When this warning is present, it takes longer to process data, and it does not handle grid cutouts correctly.  For instance, I create cutouts over the great lakes with each overpass.  For an overpass today (8 November), using the unaggregated granules, P2G V3.0 correctly created 2 cutouts only, because the  data did not cover the entire grid set.  Wheras P2G v3.0 incorrectly identified all 7 great lakes grids to be 100% covered using the aggregated data, and created GeoTIFFs for each grid, even though most are blank images.  

I have the data I am looking at on bumi:

Aggregated:
input:  `/data/users/kathys/test_data/viirs_sdr_day/2022_data`
output: `/data/users/kathys/test_data/viirs_sdr_day/output2`

Unaggregated:
input:  `/data/users/kathys/test_data/viirs_sdr_day/2022_data_puff`
output: `/data/users/kathys/test_data/viirs_sdr_day/output`

djhoese:
Closes #561. During testing it was noted that VIIRS EDR Active Fires were not as large as they used to be. This is due to the changes in nearest neighbor and that I hadn't set a default "radius_of_influence". This default set in this PR may not work for all swath and grid locations, but we'll see what happens.

Additionally this PR closes #561 where it was noticed that any NaNs existing in the lon/lat boundary coordinates were causing a lot more cases to hit the "slow path" of the boundary polygon logic. This increased processing time a lot and isn't necessary if we handle things a little smaller.


djhoese:
This depends on https://github.com/pytroll/pyresample/pull/479 being released.

djhoese:
Looks like I forgot to remove some code that isn't being used and need to add more tests for some of these edge cases. I'll do that next time I'm at my computer.

Issue #499, opened 2022-08-18
Consider adding option to ABI L1b reader for lower clipping radiances


djhoese:
I was going through old emails between some NOAA/CIMSS folks and @simonrp84 where we discussed ways to fix holes in the ABI C07 imagery that shows up in Geo2Grid/GeoSphere, but I don't think it ever resulted in an issue here. The problem we're trying to solve is that sometimes C07 has negative radiances and these result in invalid BTs when the radiances are converted to BTs. The suggestion from Mat Gunshor and Tim Schmidt was to replace the values with the minimum possible BT for a band. We determined that the best way to do that is to (quote from my response in the email thread):

```
If you assume you are looking for the lowest count where Rad is 0 or greater then you can do:

rad = count * scale_factor + add_offset

which for rad=0 and solving for count becomes:

count = -add_offset / scale_factor

If we then do a "ceiling" operation on that we should get the integer count we're looking for. And indeed when I do that with the file I've been playing with I get 197.5K. 
```

My one concern was that this assumes `add_offset` is always negative. Mat responded with:

```
add_offset will not be positive for any of the GOES-R series ABI (GOES-16 through -19) because of the design to allow for negative radiances.  I can't be sure that's true for any other instruments though. 
```

I think this behavior could be added to Satpy with an additional keyword argument, but in Satpy it would probably not be the default given how much we're "fudging" the data. I'm also not sure if it should be done for radiances or only during the conversion to BT.

djhoese:
This is now available as a reader keyword argument now that https://github.com/pytroll/satpy/pull/2499 has been merged.

djhoese:
@kathys Let's talk about this in our meeting today. FYI to @graemely, this feature as a default would fix the green "speckling" seen in GeoSphere's night microphysics composites.

djhoese:
It took me forever to find this message from @simonrp84 on the pytroll slack about the speckling in the night microphysics ABI images so I'm putting it here for future reference. This is about how even after the feature mentioned in this issue was added and tested C07-based night microphysics images still had speckling and there doesn't seem to be a reasonable way to get rid of it.

> OK, looking at the channel data, the green spots are not due to negative radiances - they're due to the quantisation applied either in the onboard ADC or via the ground segment.
[...]

PR #637, opened 2023-09-11
Add `--clip-negative-radiances` flag to ABI L1b reader

djhoese:
This flag (on by default) removes negative radiances and therefore NaN brightness temperatures. These show up in RGBs that use them, but does not solve what I had hoped it would solve as described in #499. The night_microphysics's use of C07 and the green pixels that show up are inevitable due to how C07 works. Not much we can do about that. This PR does still fix some black pixels in the RGB though.

### DIFF
--- a/polar2grid/core/script_utils.py
+++ b/polar2grid/core/script_utils.py
@@ -46,11 +46,6 @@ import os
 import sys
 from collections import defaultdict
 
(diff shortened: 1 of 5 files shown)


djhoese:
Should have mentioned:

Closes #499

PR #640, opened 2023-09-12
Add normalized burn ratio composite (nbr) for VIIRS

### LINKED ISSUE #542: Add Normalized Burn Ratio (NBR) composite example
Sam Batzli would like to use an NBR composite. Rather than having to compute it from floating point or 16-bit geotiffs, it may be better to compute it as a Satpy composite by configuring it in a YAML configuration file. Here's my attempt.

Definition: `nbr = (nir - swir) / (nir + swir)`

YAML to put in `etc/composites/viirs.yaml`:

```yaml
  nbr:
    compositor: !!python/name:satpy.composites.viirs.GenericCompositor
    prerequisites:
      - compositor: !!python/name:satpy.composites.RatioCompositor
        prerequisites:
          - compositor: !!python/name:satpy.composites.DifferenceCompositor
            prerequisites:
              - name: M11
              - name: M07
          - compositor: !!python/name:satpy.composites.SumCompositor
            prerequisites:
              - name: M11
              - name: M07
    standard_name: normalized_burn_ratio
```

Produces:

image

Note: This is dynamically linearly scaled. I'm sure there is something else people would expect.

djhoese:
Closes #542 

Waiting on comment from Sam Batzli on if there is a good default range (beyond -1 to 1) or a good default colormap.


djhoese:
I talked with Sam on slack and -1 to 1 is the expected range. He did ask that I change the use of M07 for the higher resolution I02 so I've done that in my most recent commit.

PR #653, opened 2023-11-21
Add initial FCI Level 1c NetCDF reader

djhoese:


djhoese:
There are a lot of optimizations and fixes being done on the Satpy and other pytroll package sides for this. So while this reader wrapper can be merged if/when tests pass the actual Geo2Grid release that includes it should not be done (or at least advertised) until these upstream changes have been merged and released.

These changes include:

1. Caching and other optimizations to loading/reading the files from disk.
2. Various fixes to preserve 32-bit floats and avoid upcasting to 64-bit floats which is a waste of memory.
3. Fixes for issues with some RGB composites in recent versions of Satpy.

PR #668, opened 2023-12-11
Remove unnecessary custom MiRS coefficients

djhoese:


djhoese:
Depends on https://github.com/pytroll/satpy/pull/2679

Issue #728, opened 2025-01-07
Re-add shell scripts from python package installation


djhoese:
In older versions of setuptools we could copy our bash scripts (ex. `polar2grid.sh`, `geo2grid.sh`, etc) to the `bin/` directory when someone installed the *python* package in their Python environment. This worked really well because it meant the same instructions for the tarball/bundle distribution of Polar2Grid/Geo2Grid would work for a pure-python installation.

Newer versions of setuptools have dropped support for this so in the interest of time I've just been telling users encountering this to use `polar2grid` and `geo2grid` wrapper python scripts that are installed (these are "entry points" in python packaging terms) in `bin/`.

It is my understanding that other packaging tools like poetry allow for shell script copying as I would prefer and like we used to have. I'd prefer to use `hatchling` if I can figure out how to do this, but I'm willing to try other build systems or do more hacky workarounds to get this functionality restored.

djhoese:
This should be fixed as part of #729.

PR #729, opened 2025-01-13
Update package build system to hatchling

djhoese:
This is my initial attempt to switch the python package building system from setuptools (old, but classic library) to hatchling (new hotness). My hope is that besides using more modern tooling, I'm hoping I'll be able to restore the bundled bash scripts (see #728).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,39 +52,16 @@ nosetests.xml
 *.out
 *.app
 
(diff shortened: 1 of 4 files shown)


djhoese:
I think I've covered everything here and this closes #728 from what I can tell. I'll merge this and verify the resulting software bundle that's created.

PR #748, opened 2025-04-24
Add updated AMSR2 GAASP enhancements

### LINKED ISSUE #742: Add P2G Support for CSPP AMSR2 GAASP products
This is a request to add support for the following CSPP AMSR2 GAASP Level 2 microwave products, which can be found here - `bumi:/data/users/kathys/test_data/amsr2/hampton/level2` and `bumi:/data/users/kathys/test_data/amsr2/miami/level2`

Ocean Products:  File Prefix - `AMSR2-OCEAN_*.nc`
       `double SST` - Hoping we can use similar scaling/color maps as ACSPO.
        `double WSPD`  -  This is a very popular product.  I am thinking that we should just use a simple rainbow color map to start with and let it scale from min to max.  I might write this up in the example section and then I can show people how they can scale it and add a colormap.
        `float Rain_Rate`  -  Didn't we scale and have a color map for the MiRS Rain Rate product?  I would be nice if we can reuse that.

Land Products:  `AMSR2-SOIL_*.nc`
        `float Soil_Moisture`  -  This is a percent product, based upon how much moisture the soil can hold.  So the values go from 0. to 1., and I would use a rainbow colormap.   I would also use the int `int Soil_Moisture_QA` and only include as part of the image values where `Byte 1: bit 0: soil moisture quality good is equal to 1`

Seaice Products:  `AMSR2-SEAICE_*.nc`     There is a Northern and a Southern Hemisphere product.  Both are on some kind of fixed grid with the pole at the center of the grid.  
         `int NASA_Team_2_Ice_Concentration`   The values go from 0 to 100.   I might have a color map for this one.  I need to check. Here is an image that I made from the hampton data set.

Image

Precipitation Products:  `AMSR2-PRECIP_*.nc`     
           `float Rain_Rate` 
            After looking closely at this product, it is the exact same array as found in the Ocean Products file.  So same info as that.

Snow Products:  `AMSR2-SNOW_*.nc`
          `float Snow_Depth`  in cm.  I do not know if this matters or not, but the product is chunked. `:_ChunkSizes = 225U, 243U; // uint`

I put some suggestions for color maps, but let me work on that.  I have a couple from the snow/ice SSEC team that we might be able to use.

djhoese:
Closes #742

## Miami Wind Speed

<details>

image

</details>

## TPW

<details>

image

</details>

## SWE

<details>

image

</details>

## SST

<details>

image

</details>

## Soil Moisture (no QC)

<details>

image

</details>

## Snow_Depth

<details>

image

</details>

## Snow_Cover

<details>

image

</details>

## Rain Rate

<details>

image

</details>

## CLW

<details>

image

</details>

### DIFF
--- a/polar2grid/etc/enhancements/amsr2.yaml
+++ b/polar2grid/etc/enhancements/amsr2.yaml
@@ -1,0 +1,144 @@
+enhancements:
+  gaasp_clw:
+    name: CLW
+    sensor: amsr2
+    operations:
+      - name: linear_stretch
+        method: !!python/name:satpy.enhancements.stretch
+        kwargs: {stretch: 'crude', min_stretch: 0.0, max_stretch: 0.7}
+  gaasp_sst:
+    name: SST
+    sensor: amsr2
+    operations:
+      - name: linear_stretch
+        method: !!python/name:satpy.enhancements.stretch
+        kwargs: {stretch: 'crude', min_stretch: 267.317, max_stretch: 309.816}
+  gaasp_tpw:
+    name: TPW
+    sensor: amsr2
+    operations:
+      - name: colorize
+        method: !!python/name:satpy.enhancements.colorize
+        kwargs:
+          palettes:
+            - filename: colormaps/idl_rainbow.cmap
+              min_value: 0
+              max_value: 75
+  gaasp_wspd:
+    name: WSPD
+    sensor: amsr2
+    operations:
+      - name: linear_stretch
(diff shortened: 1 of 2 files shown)


kathys:
Thumbs up @djhoese .

djhoese:
Merging this as I think, like all of our updates for this round, we just need to do some real comparisons and fine adjustments of limits but the basic functionality and basic appearance of these products looks good.

PR #789, opened 2026-01-08
Add MiRS rain rate colormap for MiRS and GAASP rain rate

### LINKED ISSUE #785: GAASP rain rate does not match MiRS rain rate colormapping
As identified by @DougASchu in an email thread:

> The GAASP Rain Rate has 0 as purple, instead of transparent like the MiRS Rain Rate.

Identified in P2G 3.2rc1.

djhoese:
Closes #785 

Not sure if I just have a bad dataset or what but here is the old colormap for a dataset I have and second is after this PR. Note both have `--fill-value 0`.

image

image

This is from an old GAASP file.

```
AMSR2-PRECIP_v2r2_GW1_s202503241851510_e202503241903060_c202503241918300.nc
```


kathys:
Yes, this is strange how much it changed.  So you changed the GAASP Rain Rate range, so you would expect to see more detail in the output product.  I guess I will look closely at the output once I put it into my operations.  I like the new colormap better than the old.  But you are right, we want to make sure it is correct.

kathys (approved):

PR #803, opened 2026-04-01
Add SCMI lettered grid images to docs

djhoese:
Also updates references to crefl HDF4 files. Now that SSEC doesn't have personal websites I can't store these files under `~davidh` so I put the PNG images here in the repository and pointed to the HDF4 files on CSPP's FTP location.


djhoese:
Note the HDF4 changes were also needed for Satpy: https://github.com/pytroll/satpy/pull/3365

PR #810, opened 2026-04-24
Add initial VIIRS S3 search script

djhoese:
This PR adds a new utility script called `search_noaa_s3_viirs.sh` which allows the user to specify the satellite, bands, and start/end time that they wish to pull from NOAA's AWS S3 buckets. Satellite can be snpp, n20, or n21. The way this works is:

1. Iterate over hour between start and end, keeping in mind that we want the granule to cover these times and may need to search earlier than start time to find a granule.
2. For each hour, generate a glob pattern for S3.
3. Glob the S3 bucket.
4. For each result, do a finer grained search on the time of that file to make sure it over laps with the times we're concerned about.

My script includes an optimization using the `s3fs` library. Instead of using `glob` directly it uses the "find" functionality to search for all objects under a path that match a particular "prefix". For many glob patterns (times or bands) this proves to be much faster. The regular `glob` functionality requires listing and searching through all objects in the bucket.

This example call should work:

```bash
polar2grid.sh -r viirs_sdr -w geotiff -vvv -f @<(search_noaa_s3_viirs.sh -s n20 -b I04 -b GITCO --start-time 2026-04-23T12:00:00 --end-time 2026-04-23T12:01:00)
```

You can also specify `-b ALL` to do all bands and geolocation files. Note NOAA buckets don't have up to date non-terrain corrected files.

### DIFF
--- a/create_conda_software_bundle.sh
+++ b/create_conda_software_bundle.sh
@@ -94,6 +94,7 @@ if [[ $PROJECT == "P2G" ]]; then
     cp $BASE_P2G_DIR/NEWS.rst $SB_NAME/RELEASE_NOTES.txt || oops "Couldn't copy release notes to destination directory"
 else
     rm -f bin/polar2grid.sh
+    rm -f bin/search_noaa_s3_viirs.sh
     rm -f bin/POLAR2GRID_README.txt
     mv bin/GEO2GRID_README.txt README.txt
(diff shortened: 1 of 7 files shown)


djhoese:
I should have added to the above description:

The magic in the example call is:

* `@` tells the python code to take in the argument following it as a file to read more arguments from
* `<(command)` runs the command and pipes its output to a temporary file and outputs the path to that file into the command. Python is reading this file.

This PR also allows for `search_noaa_s3_viirs.sh ... | polar2grid.sh ... -f -` where `-` represents stdin.

Issue #816, opened 2026-06-03
Add OMPS EDR reader with default products


djhoese:
* `ColumnAmountO3`
* `AerosolIndex`
* `Reflectivity331`

Issue #817, opened 2026-06-03
Add OMPS EDR product AWIPS file generation support


kathys:
Please add the option to use the `awips_tiled` writer for the 3 standard default OMPS EDR products - See #816 .

PR #818, opened 2026-06-04
Add 'omps_edr' reader

djhoese:
Depends on https://github.com/pytroll/satpy/pull/3414 being merged before a tarball is built.


djhoese:
This PR will close #816 and #817. @kathys I tested the AWIPS tile generation and it seems to work fine once the Satpy PR mentioned above is merged. There was one small issue with `valid_range` that is fixed in that Satpy PR.

PR #832, opened 2026-08-24
Add initial 'metimage_l1b_nc' reader

djhoese:
Written by Claude.


djhoese:
There is an optional orthorectify option in Satpy. I'm still figuring out if that is useful or recommended.

kathys (commented):
Just one comment.  Can you please add support for `false_color` RGBs too?

djhoese:
Natural color (most of this recipe comes from Satpy I think):

image

True color:

image

djhoese:
I started looking into why the true color looks darker than other instruments (ex. VIIRS) and the conclusion is that the blue channel is far enough away wavelength-wise that it isn't able to produce the same type of image. There are other small things like VIIRS having higher resolution versions of the channel so ratio sharpening is being performed and VIIRS in Polar2Grid also uses CREFL atmospheric correction versus MetImage's pyspectral rayleigh correction, but those contributed much less to the difference between the two true colors.

djhoese:
Note pyspectral needs [this PR](https://github.com/pytroll/pyspectral/pull/292) to avoid a lot of warnings about the instrument name not being found.

djhoese:
@kathys Do you know if Liam has tried EWA/fornav resampling on metimage data? I'm curious if he has guesses for the best flags (-d/-D). Right now I'm using 40/2 which matches VIIRS. I would ask Andrea from EUMETSAT but he just left for vacation.

kathys:
Dave, I asked Liam about the best EWA options for MetImage data, and he said he has not tried to make an image from the data yet.  But he said Douglas has.  @DougASchu  can you provide us with any info on the EWA resampling options that you are using?

DougASchu:
@kathys I've been using the Satpy METimage reader built-in to P2G v3.2 without any specific adjustments to the resampling, so whatever the default resampling is from Satpy/P2G is what we've been using to create GeoTIFFs and then downstream PNGs/JPGs/webp images.  For example:

sga1_vii_vii_668_20260825_160042_central_north_america

Using a command like: 
`polar2grid.sh -r vii_l1b_nc -w geotiff -p vii_668 --fill-value 0 --keep-palette -g central_north_america --grid-configs /home/dbvm/dbvm/scripts/grid_configs/psas_regional.yaml --grid-coverage=0.01 --extra-config-path /home/dbvm/dbvm/scripts/p2g_configs/conus --extra-config-path /home/dbvm/dbvm/scripts/p2g_configs/common  --num-workers 16 -f input`

@djhoese I also just wanted to make a comment on the METimage true color discussion above.  @spruceboy and I noticed this true color mismatch, and I came up with a set of tweaks to treat the METimage bands like their MODIS counterparts in the CREFL code, which isn't a 1-to-1 match with VIIRS CREFL true color but gets fairly close to MODIS CREFL true color (see examples below).  Certainly it reduces the "yellow skew" effect.  I could share that if you are interested.  I wouldn't call it validated in any way, just something to make the output more pleasant to look at.

sga1_vii_true_color_20260821_154330_MADISON
noaa21_viirs_true_color_20260821_191157_MADISON
aqua_modis_true_color_20260821_210523_MADISON

djhoese:
Thanks @DougASchu. I think I'm going to avoid CREFL for now just with how much people want to get this reader and data processed. I'm getting a pull request ready for Satpy that should improve dask chunking for metimage a bit, but I still need to get some thing finalized.

I asked some EUMETSAT folks and they said they didn't know of any effort to make the MetImage true color more like VIIRS. I do notice in your images Douglas that it is much closer to MODIS than I expected.

Edit: For my future reference, a chunk size in P2G of 1916 seems to perform best with these planned changes.

djhoese:
Not sure if it is MetImage or METImage or METimage...I'll merge this when CI passes and then fix the naming later.

djhoese:
Ah I also never finalized the orthorectify decision and if that was useful or preferred.

PR #836, opened 2026-09-02
Add new --filter-negative-so2 flag 

### LINKED ISSUE #820: Add support for OMPS SO2 Products
Please add a Polar2Grid Reader for the JPSS OMPS V8TOS* prefixed NetCDF files.  I would like the following arrays to be the standard ones that are supported.

`ColumnAmountOzone`  This is the same array as the one supported in OMPS Total Ozone V8TOZ files, including having the option to use the  `ErrorFlag` array.

---------------------------------------------------------------------------------------------------------

I would also like these 4 arrays to be supported. They should be handled in the same way - each represent the amount of SO2 for for different height distributions:
`S_COLUMNAMOUNTSO2_PBL`
`S_COLUMNAMOUNTSO2_TRL`
`S_COLUMNAMOUNTSO2_TRM`
`S_COLUMNAMOUNTSO2_STL`

The arrays are floats.

Values are in Dobson units.  I would like the to use the same color table as is used by JSTAR Mapper for these 4  arrays, which Gemini says is this:
`# OMPS SO2 Color Map LUT `
0,0,0,0,0
1,0,204,255,255
50,0,255,0,255
100,255,255,0,255
150,255,153,0,255
200,255,0,0,255
230,204,0,204,255
255,0,0,0,255

The SO2 values for this color table go from 0-10., with everything above 10 to be pinned to the last color (red), as shown in the JSTAR Mapper website:

Image

---------------------------------------------------------------------------------------------------------

And then the following 3 arrays should be treated the same as the ColumnAmountOzone, because they are adjusted Total Column Ozone based upon the amount of SO2.
`s_STLO3`
`s_TRLO3`
`s_TRMO3`
They are in Dobson units.  So the same color table should be used (Rainbow) as is used for ColumnAmountOzone.

I will ask the algorithm developer about Quality Flags for these other products.  I forgot to do that.

I put sample files for the 3 JPSS instruments at:

`bumi:/data/users/kathys/test_data/omps/j02/tos`
`bumi:/data/users/kathys/test_data/omps/j01/tos`
`bumi:/data/users/kathys/test_data/omps/npp/tos`

djhoese:
Also rename `--filter-o3` flag to `--filter-by-error-flag`.

The new flag is on by default, error flag flag is still off by default.

Should close #820 but we could also wait for @kathys to have time to test it more thoroughly. This PR depends on https://github.com/pytroll/satpy/pull/3455 being merged first for the new keyword argument to be available.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -16,7 +16,7 @@ import os
 import sys
 import urllib.request
 import warnings
-from datetime import datetime, UTC
+from datetime import UTC, datetime
 from shutil import copyfileobj
 
 import yaml
--- a/doc/source/readers/omps_edr.rst
+++ b/doc/source/readers/omps_edr.rst
@@ -23,7 +23,9 @@ Examples:
 
     polar2grid.sh -r omps_edr -w geotiff -f /omps/edr/
 
-    polar2grid.sh -r omps_edr -w geotiff -p ColumnAmountO3 --filter-o3 -f ../omps/V8TOZ-EDR_v4r3_j01_*.nc
+    polar2grid.sh -r omps_edr -w geotiff -p ColumnAmountO3 --filter-by-error-flag -f ../omps/V8TOZ-EDR_v4r3_j01_*.nc
+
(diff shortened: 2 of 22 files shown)


djhoese:
The Satpy PR isn't merged yet, but I have some other maintenance I want to do on this project so I'm going to merge this as it has some changes I want.

PR #843, opened 2026-09-17
Update 'viirs_edr' docs to mention new LST name (was VLST)

djhoese:
The new variable name in the v2r2 files is "LST". The old name was "VLST". Depends on https://github.com/pytroll/satpy/pull/3474


djhoese:
FYI @kathys, users can still do VLST with old files but our documentation will say "LST" to match the new files. I do not have anything in Satpy to map VLST to LST or vice versa.

kathys:
Sounds good @djhoese

